PR #35, opened 2019-09-25
Enhance/remove unneeded props

3nvi:
### Background

As the PantherUI got developed, small fixes on pounce were made in order to accommodate both needs of the UI repo, as well as issues that were caused by small bugs in Pounce.

Instead of creating a pull request for each bugfix, I chose to put them all into a branch in order to speed up development of the PantherUI. Now its time for those changes to be properly added. 

### Changes

- Table accepts now a `renderPlaceholder` prop
- All buttons have `type="button"` by default now
- Added necessary `e.preventDefault()` to some events 
- Added `useTheme` hook
- Properly exposing typings for theme
- Padding fixes here & there


src/components/Table.md
```diff
@@ -671,11 +671,7 @@ const Example = () => {
 
   return (
     <Card width="100%">
-      <Table
-        items={items.slice(startIndex, endIndex + 1)}
-        getItemKey={item => item.id}
-        columns={columns}
-      />
+      <Table items={[]} getItemKey={item => item.id} columns={columns} />
```
drixta: Why are we not passing in the `items` to the Table component here?

3nvi: Because it's a markdown example where the user simply adds items instead of selecting from existing ones

PR #37, opened 2019-10-16
Feature/emotion

3nvi:
### Background

This PR replaces `styled-components` with `emotion`. The sole reason for doing that is to utilise the `shouldForwardProp` option that `emotion` supports, which allows us to not forward un-needed props to the DOM.

This removes all React warnings during development (with regards to un-recognized props on DOM elements), as well as make the overall markup more clean. 

### Changes

- Replace `styled-components` with `emotion` in a 1:1 mapping

### Testing

- If the migration is successful, everything should work as before


jacknagz:
I'm glad we're adding more emotion into our UI

3nvi:
> I'm glad we're adding more emotion into our UI

I'll change our description to "How you feel is our priority"

PR #39, opened 2019-10-18
CI: Update CircleCI config

3nvi:
### Background

Makes sure to automatically create versions on master pushes 

### Changes

- Add semantic-release on master pushes

### Testing

- Merging this to master should automatically create a new release ^^

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
   release:
     executor: node
     steps:
-      - build
+      - setup
       - run:
           name: Create a new release with proper changelog
           command: npx install semantic-release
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -43,9 +43,10 @@ jobs:
   release:
     executor: node
     steps:
+      - setup
       - run:
           name: Create a new release with proper changelog
-          command: npx install semantic-release
+          command: npx semantic-release
```
jacknagz: Can you explain this command a little more? I'm not sure what it does

3nvi: This uses the ENV variables from CircleCI to perform a package release on Github/NPM. We currently have the proper authentication env vars setup on CircleCI and that's how it has the permissions to work

PR #62, opened 2020-04-27
Fix for ComboBox for value === false

alexmylonas:
### Background

ComboBox seems to not support [`true`,`false`] values. We use this on filtering policies `Auto-remediation Status` and selecting the `false` value has no effect!
 
### Changes

- Updated check value on ComboBox

### DIFF
--- a/src/components/Combobox/Combobox.tsx
+++ b/src/components/Combobox/Combobox.tsx
@@ -97,7 +97,7 @@ function Combobox<ItemShape>({
   maxResults,
 }: ComboboxProps<ItemShape>): React.ReactElement<ComboboxProps<ItemShape>> {
   // convert item to a string with a fallback of empty string
-  const safeItemToString = (item: ItemShape | null) => (item ? itemToString(item) : '');
+  const safeItemToString = (item: ItemShape | null) => (item != undefined ? itemToString(item) : '');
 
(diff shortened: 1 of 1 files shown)


src/components/Combobox/Combobox.tsx
```diff
@@ -163,7 +163,7 @@ function Combobox<ItemShape>({
               'aria-readonly': true,
             }),
             ...(searchable && {
-              placeholder: value ? itemToString(value) : inputProps.placeholder,
+              placeholder: value !== undefined ? safeItemToString(value) : inputProps.placeholder,
```
3nvi, line 166: Gotcha! Interesting, that was definitely a bug, but I'm not 100% sure if that fixes it.

With your change, if value is `false`, then in L166 it would enter `safeItemToString(value)`, where the related code 

```ts
  const safeItemToString = (item: ItemShape | null) => (item ? itemToString(item) : '');
```

would return into the `placeholder` getting the value `''`. Wouldn't we want to show `"false"`?


I think the solution would be to keep using `itemToString` and not use `safeItemToString`

alexmylonas: You are right and also notice that if I dont update the `safeItemToString` it would display the value but dont handle the value anyway, so I changed it, have a look!

PR #78, opened 2020-07-07
Fix safari intersection observer polyfill dependency

vorillaz:
### Background

As `react-use-measure` [docs](https://github.com/react-spring/react-use-measure/#%EF%B8%8F-notes) suggest the hook requires a polyfill in order to properly work with Safari. The repo was not properly working and the webpack error tracking was complaining on startup. This PR adds the polyfill as suggested.

Screenshot 2020-07-07 at 12 37 53 PM

### Changes

- Add the [juggle/resize-observer](https://github.com/juggle/resize-observer) as suggested to `package.json` and `package-lock.json`
- Patch the `useMeasure` React hook to consume the polyfill 

### Testing

- Check out the master branch, run `npm i` and `npm run start`
- Open the app in Safari (I am currently using the version 13.0.5 which does not support the [resize observer](https://caniuse.com/#feat=resizeobserver)
- Check out the PR branch, once more run run `npm i` and `npm run start`
- The Snackbar component should work as expected with no errors or warnings.
- You can also test the changes in the main `panther` repo as follows: 
  - Go to `pounce` repo and run `npm pack` a `pouncejs-0.0.0-semantically-released.tgz` should get created at the root of the project
  - Replace the actual `pouncejs` entry in your `package.json` with the absolute path targeting the tarball created in the previous step, e.g `"pouncejs": "/Users/theodore/Desktop/pounce/pouncejs-0.0.0-semantically-released.tgz"`
  - Run  `npm i` and `npm run start` and open the panther app, everything should be working as expected with no errors or warnings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1515,6 +1515,11 @@
       "resolved": "https://registry.npmjs.org/@emotion/weak-memoize/-/weak-memoize-0.2.5.tgz",
       "integrity": "sha512-6U71C2Wp7r5XtFtQzYrW5iKFT67OixrSxjI4MptCHzdSVlgabczzqLe0ZSgnub/5Kp4hSbpDB1tMytZY9pwxxA=="
     },
+    "@juggle/resize-observer": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/@juggle/resize-observer/-/resize-observer-3.2.0.tgz",
+      "integrity": "sha512-fsLxt0CHx2HCV9EL8lDoVkwHffsA0snUpddYjdLyXcG5E41xaamn9ZyQqOE9TUJdrRlH8/hjIf+UdOdDeKCUgg=="
+    },
     "@marionebl/sander": {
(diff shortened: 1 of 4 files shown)


vorillaz:
Thanks for the feedback @3nvi. I had similar concerns about the polyfill's size, the other way around would be checking the availability of the API in the browser but we would pollute the `window` object.

As per the two comments about the CI are there any actions I should make? Maybe opening a new PR or fixing the failing build in this one?

3nvi:
> Thanks for the feedback @3nvi. I had similar concerns about the polyfill's size, the other way around would be checking the availability of the API in the browser but we would pollute the `window` object.
> 
> As per the two comments about the CI are there any actions I should make? Maybe opening a new PR or fixing the failing build in this one?

The `audit` can't be fixed until the upstream (lodash) handles it.

The `build` is interesting that it's failing. I thought it was an issue with the current code, but it's unrelated. 

Feel free to merge and I'll push a fix for those

PR #81, opened 2020-07-09
fix: issue with JSDom and TextAreaAutosize

3nvi:
### Background

Having `border: none` messed up with height measurements  in jsdom. Making it be `0` fixes that.

### Changes

- Change `border: none` to `border: 0`

### Testing

- N/A


vorillaz:
LGTM, why are you referring to JSDom though? Is there any JSDom checking in place yet?

3nvi:
> LGTM, why are you referring to JSDom though? Is there any JSDom checking in place yet?

When I import  it  into `panther` and attempt to use it in a test,  that's  when JSDom assigns `NaN` height  to the `textarea` element ;)

PR #92, opened 2020-08-04
feat: add styling to tabs

3nvi:
### Background

Lots of designs have tabs that have the same styling applied to them. This PR adds this styling on pounce so that  apps like Panther can use them without any trouble.

### Screenshots
Screen Shot 2020-08-04 at 3 20 08 PM

### Changes

- Pass an `index` prop from `TabList` to  all `Tab`  components to help with styling
- Create  `useTabStyles` local hook to handle  all styling
- Use hook in `Tab` components

### Testing

- Testing steps

### DIFF
--- a/src/components/Tabs/useTabStyles.tsx
+++ b/src/components/Tabs/useTabStyles.tsx
@@ -1,0 +1,37 @@
+import { useTabsContext } from '@reach/tabs';
+import { AbstractButtonProps } from '../AbstractButton';
+
+interface UseTabStylesProps {
+  index: number;
+}
+
+const useTabStyles = ({ index }: UseTabStylesProps): Partial<AbstractButtonProps> => {
+  const { focusedIndex, selectedIndex } = useTabsContext();
+
+  const selectedColor = 'blue-400';
+  const focusedColor = 'navyblue-300';
+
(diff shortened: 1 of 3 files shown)


src/components/Tabs/useTabStyles.tsx
```diff
@@ -0,0 +1,37 @@
...
+  let borderColor: AbstractButtonProps['borderColor'];
+  if (isSelected) {
+    borderColor = selectedColor;
+  } else if (isFocused) {
+    borderColor = focusedColor;
+  } else {
+    borderColor = 'transparent';
+  }
+
+  return {
```
vorillaz, line 26: I guess that's intended but IMHO tabs should get visual impact for keyboard navigation.

3nvi: They do!

that's what `isSelected` and `isFocused` handle through the border. 

On tabs where the navigation instantly changes the tab,  you get immediate feedback.

On tabs where it requires pressing "Enter"  to "go" to the tab, then  you get feedback for your selection. 

All that is handled internally through reach-ui and presented through a border. I just didn't want the outline, since there was already another visual indicator present

PR #93, opened 2020-08-04
Fix issues with tab focuses

3nvi:
### Background

Currently tabs are not "outlined" when a user users the `<tab>` key. At the same time, when you tab, the entire tabpanel gets focused which you probably don't want that.

This PR fixes that by making sure focus is handled correctly. Also it  adds a missing border below the tabs (SS)

### Screenshots

Problematic focusing can be seen here

Screen Shot 2020-08-04 at 5 09 11 PM

A new border has been added 

Screen Shot 2020-08-04 at 5 15 15 PM

### Changes

- Add border
- Re-instate tab outline
- Remove tab panel focus

### Testing

- Visual

### DIFF
--- a/src/components/Tabs/TabPanel.tsx
+++ b/src/components/Tabs/TabPanel.tsx
@@ -1,6 +1,7 @@
-import { TabPanel as ReachTabPanel, TabPanelProps as ReachTabPanelProps } from '@reach/tabs';
+import React from 'react';
+import { TabPanel as ReachTabPanel } from '@reach/tabs';
+import Box from '../Box';
 
-export type TabPanelProps = ReachTabPanelProps;
-export const TabPanel = ReachTabPanel;
+const TabPanel: React.FC = props => <Box as={ReachTabPanel} tabIndex={-1} {...props} />;
 
 export default TabPanel;
--- a/src/components/Tabs/useTabStyles.tsx
+++ b/src/components/Tabs/useTabStyles.tsx
@@ -24,7 +24,7 @@ const useTabStyles = ({ index }: UseTabStylesProps): Partial<AbstractButtonProps
   }
 
   return {
-    outline: 'none',
+    zIndex: 1,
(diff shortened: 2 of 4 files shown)


src/components/Tabs/useTabStyles.tsx
```diff
@@ -24,7 +24,7 @@ const useTabStyles = ({ index }: UseTabStylesProps): Partial<AbstractButtonProps
   }
 
   return {
-    outline: 'none',
+    zIndex: 1,
```
vorillaz, line 27: I don't think that manipulating the items would do the trick.

src/components/Tabs/TabPanel.tsx
```diff
@@ -1,6 +1,7 @@
-import { TabPanel as ReachTabPanel, TabPanelProps as ReachTabPanelProps } from '@reach/tabs';
+import React from 'react';
+import { TabPanel as ReachTabPanel } from '@reach/tabs';
+import Box from '../Box';
 
-export type TabPanelProps = ReachTabPanelProps;
-export const TabPanel = ReachTabPanel;
+const TabPanel: React.FC = props => <Box as={ReachTabPanel} tabIndex={-1} {...props} />;
```
vorillaz, line 5: That's breaking a screen reader as we need to ensure that the content is stacked properly.

src/components/Tabs/useTabStyles.tsx
3nvi: the `zIndex` is for an unrelated issue (where the border of the `Tab` should be above the pseudoelement of the `TabList`)

src/components/Tabs/TabPanel.tsx
3nvi: how can I tell the screenreader "don't focus on the tabpanel, but on its content"?

The reason this focus exissts is because the lib that we are using had a `tabIndex=1` on a `div`

PR #97, opened 2020-08-05
Feat test setup

vorillaz:
### Background

Implementing the first Jest setup for pounce 🐯 

### Changes

- Added Jest, TS support, snapshot testing support and Jest-Axe for A11Y testing.
- Added a few tests for demo purposes.
- Fixed a tiny A11Y bug for the `<Badge />` component.


jest/babel.config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = api => {
```
3nvi, line 1: can we use the existing babel config or perhaps extend it? In Panther we do something similar and I'd rather have a single babel configuration

jest/config.js
```diff
@@ -0,0 +1,16 @@
+const { defaults } = require('jest-config');
+
+module.exports = {
+  rootDir: '../',
+  testMatch: ['<rootDir>/src/**/*.test.{ts,tsx}'],
+  testPathIgnorePatterns: ['/dist/', '/.github/', '/node_modules/'],
+  moduleFileExtensions: [...defaults.moduleFileExtensions, 'ts', 'tsx'],
+  setupFilesAfterEnv: ['<rootDir>/jest/setup.ts'],
+  verbose: true,
+  transform: {
```
3nvi, line 10: I think that we normally wouldn't want this explicit transform. By default jest picks up the top level babel config, so if we re-use it then we wouldn't need this manual transform 😃

jest/utils/index.ts
```diff
@@ -0,0 +1,8 @@
+import { fireEvent } from '@testing-library/react';
+export * from '@testing-library/react';
+
+export function fireClickAndMouseEvents(element: HTMLElement) {
```
3nvi, line 4: out of curiosity, wouldn't this be a double click? If you are **just** using it for the `Dropdown` component, just letting you know that the Dropdown just needs `mouseDown` 😅 

Also, I'd put  this in a `helpers` file (in my mind `index.ts` files only export)

package.json
```diff
@@ -89,15 +92,23 @@
     "@commitlint/cli": "^8.3.5",
     "@commitlint/config-conventional": "^8.3.4",
     "@emotion/babel-preset-css-prop": "^11.0.0-next.10",
+    "@emotion/cache": "^10.0.29",
```
3nvi, line 95: how come we installed emotion 10? 

(both cache & core)

src/components/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <>
+      <Badge color="red-300">INFO</Badge>
+      <Badge color="teal-500">LOW</Badge>
+      <Badge color="blue-400">WHATEVER</Badge>
+    </>
+  );
+  expect(container).toMatchSnapshot();
+});
+
+it('aligns with a11y', async () => {
```
3nvi, line 19: nice!

package.json
```diff
@@ -89,15 +92,23 @@
...
+    "@emotion/core": "^10.0.28",
+    "@testing-library/jest-dom": "^5.11.1",
+    "@testing-library/react": "^10.4.7",
     "@typescript-eslint/eslint-plugin": "^3.1.0",
     "@typescript-eslint/parser": "^3.1.0",
+    "babel-jest": "^26.1.0",
```
3nvi, line 101: might not need it if we don't use a jest transform

3nvi:
Before we merge it, can you make sure that:

- build is passing (fix `jest-axe` types, check the details from CircleCI 😉 )
- add another step  to circle CI so that the tests run everytime :)

jest/utils/index.ts
vorillaz: That part was intended and picked from Reach UI as the API requires a bit of tweaking in order to work.
https://github.com/reach/reach-ui/blob/develop/packages/menu-button/__tests__/menu-button.test.tsx#L2
I'll fix the helpers :)

3nvi: Oh cool! Ok then 😄

PR #107, opened 2020-08-25
feat: add icon support for TextInput component

vorillaz:
### Background

Add icon support for `<TextInput />` component.

### Changes

- Add icon renderer with the same principles as the 

### Testing

- Added snapshot coverage.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import { NativeAttributes } from '../Box';
 import { slugify } from '../../utils/helpers';
 import { InputControl, InputElement, InputLabel } from '../utils/Input';
+import Icon, { IconProps } from '../Icon';
+import Flex from '../Flex';
 
 export type TextInputProps = NativeAttributes<'input'> & {
   /** The label that is associated with this input */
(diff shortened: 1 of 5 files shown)


3nvi:
@vorillaz Cool!

Should we perhaps added  for the `Textarea` component as well? Lastly, currently, if you "mouse over" the icon you won't be able to "click" it to go on the underlying  `input`. Should we perhaps add it via a `position: absolute` and simply add more padding on the right side of the `<InputElement>`?

This way the input can be clickable end-to-end and clicking the icon will still put the focus on the input (will also remove the need for a wrapping `<Flex>` component). What do you think?

vorillaz:
Nice catch @3nvi , updated the PR with the icon placed in absolute position.
I'll ask for a design with the textarea and I'll come back to you ASAP. 
Up until then, I think we can merge this one.

3nvi:
> updated the PR with the icon placed in absolute position.

@vorillaz is there a chance you forgot to commit your changes? 😆

vorillaz:
@3nvi 😨  Done!

src/components/TextInput/TextInput.tsx
```diff
@@ -15,29 +17,38 @@ export type TextInputProps = NativeAttributes<'input'> & {
...
-      </InputLabel>
+      <Flex align="center" position="relative">
+        <InputElement
+          as="input"
+          type="text"
+          id={identifier}
+          name={name}
+          value={value}
+          truncated
+          pr={icon && 40}
```
3nvi, line 41: For future reference, we use a `4px` scale, thus `10`  = `40px`. 

Now, the scale goes from `0` up to `10`. Anything above it is **automatically** translated into `px` as a fallback. Thus in your case, using either `10` or `40` would be the same. This is because:

- 0 = 0
- 1 = 4px
- 2 = 8px
....
- 10 = 40px

**SCALE STOPS**

- 11 = 11px
- 12 = 12px
....
- 40 = 40px

Similarly:

src/components/TextInput/TextInput.tsx
```diff
@@ -15,29 +17,38 @@ export type TextInputProps = NativeAttributes<'input'> & {
...
+          position="relative"
```
3nvi, line 42: not use if `position: relative` is needed

3nvi (approved):
LGTM. 

Just a minor comment

src/components/TextInput/TextInput.tsx
vorillaz: Position relative is needed as `z-index` is used in order to stack up and send the icon to the "background".

PR #108, opened 2020-08-25
test: update snapshot tests for Tabs

vorillaz:
### Background

Update failing snapshot tests.


3nvi (approved):

3nvi:
Minor comment. Use `test: ...` as the commit message when updating stuff that **won't be shipped** to the user. Generally, `fix: ...` creates a patch version (which we don't need when updating a snapshot test 😄 )

PR #109, opened 2020-08-25
chore: extend ButtonProps to assign styled system values

vorillaz:
### Background

Extend ButtonProps to assign styled system values as we need to use some spacing across the Buttons.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,11 +1,15 @@
 import * as React from 'react';
 import AbstractButton from '../AbstractButton';
 import { BoxProps, NativeAttributes } from '../Box';
+import { SystemProps } from '../Box/system';
 import useButtonStyles from './useButtonStyles';
 import Spinner from '../Spinner';
 import Icon, { IconProps } from '../Icon';
 
-export interface ButtonProps extends NativeAttributes<'button'>, Pick<BoxProps, 'as' | 'to'> {
+export interface ButtonProps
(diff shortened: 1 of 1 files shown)


3nvi:
I actually have originally configured it like that intentionally. The purpose of spacing falls on the wrapper elements and not the buttons themselves.

For example, to add spacing between buttons, you would do:

```tsx
<Flex spacing={4}>
  <Button1 />
  <Button2 />
</Flex>
```

I'm trying to limit the API of some components in order to make them less prone to "overriding" and errors. Disabling `SystemProps` disallows you to accidentally modify the padding or the border radius of this component. On the contary, in a `<Text>` or a `<Box>` component you can do whatever you want .

Thoughts?

vorillaz:
Agreed, closing this one without merging. Thanks for the comment.

3nvi:
> Hm, that works when everything is enclosed inside a parent element, but what happens when I just wanna add some margin right?

Every time you need a margin, there is a  "layout" need,  which means that things are placed inside a container already. So far, we have built Panther without any margin on buttons, so I feel that most likely you can do the  same.

P.S. If you really  need it as a one-off, you can just `// @ts-ignore` it. In the future, I plan to also discuss the removal of such props from other components like `Icon` & `Link`. This helps make fewer mistakes, since, for example, with you change another developer might hcange the "padding" of the button and end up with something that doesn't conform to our design spec

vorillaz:
You're right, closing this one!

PR #111, opened 2020-08-25
fix: Input labels

vorillaz:
### Background

Due to the design of the DateInput component, we have some combo boxes without any labels attached. This PR allows the creation of such visual effects.

### Changes

- Introduce a new prop called `standalone` which allows the input elements to work just with the placeholder text.

### Screenshots
Screenshot 2020-08-26 at 10 24 01 AM

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -118,6 +118,7 @@ function MultiCombobox<Item>({
   disableItem = () => false,
   searchable = false,
   label = '',
+  labelHidden,
   disabled = false,
   placeholder = '',
   itemToString = item => String(item),
@@ -139,6 +140,7 @@ function MultiCombobox<Item>({
     }
   };
 
+  const itemsPt = labelHidden ? 3 : '19px';
   return (
     <Downshift<Item>
       stateReducer={stateReducer}
@@ -231,7 +233,7 @@ function MultiCombobox<Item>({
                 hidden={hidden}
               >
                 {value.length > 0 && (
-                  <Flex as="ul" wrap="wrap" pl={3} pr={10} pt="19px" pb="2px">
+                  <Flex as="ul" wrap="wrap" pl={3} pr={10} pt={itemsPt} pb="2px">
                     {value.map(selectedItem => (
                       <Tag
                         as="li"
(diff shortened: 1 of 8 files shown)


3nvi:
I'm fine with  those changes, but I feel that they should affect all related components, not selectively what you need. This means that the `label` should now be optional in:

- `TextInput`
- `TextArea`
- `NumberInput`
- `DateInput` (we can skip this one if we are going to remove it)
- `Combobox`
- `MultiCombobox`

P.S. The relate design screens have `HH` and `MM` as `<TextInput />` so you need it eitherway

vorillaz:
Ι am not sure we need to update all the related inputs at once. It makes no sense to use inputs without labels but this is a highly specific need that affects just the combo box. Take a look at the related screenshots attached.

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -18,6 +18,9 @@ export type MultiComboboxProps<T> = {
   /** The label associated with this dropdown form element */
   label: string;
 
+  /** Whether the label should get visually hifen */
+  labelHidden?: boolean;
```
3nvi, line 22: minor typo here

3nvi: nitpick: can we rename it to `hideLabel`?

src/components/utils/Menu/Menu.tsx
```diff
@@ -5,6 +5,8 @@ import Box from '../../Box';
 const AnimatedBox = animated(Box);
 
 interface MenuProps {
+  /** Whether the current menu is a standalone item without label attached **/
```
3nvi, line 8: do we need this prop here?

src/components/utils/Input/InputElement.tsx
```diff
@@ -30,16 +39,20 @@ const InputElement = React.forwardRef<HTMLInputElement | HTMLTextAreaElement, In
...
-          color: 'gray-50',
-          transition: 'opacity 200ms cubic-bezier(0.0, 0, 0.2, 1) 0ms',
-        }}
-        _focus={{
-          '::placeholder': {
-            opacity: 0.4,
-          },
-        }}
+        _placeholder={
+          !standalone && {
```
3nvi, line 43: TS might warn you here (after my PR gets merged). 

Ideally you'd  want `!standalone ? styles : {}` instead of `!standalone && styles`

3nvi (approved):
All good.

Left a  few comments :)

src/components/utils/Input/InputElement.tsx
vorillaz: Done!

src/components/utils/Menu/Menu.tsx
vorillaz: You're right.

src/components/MultiCombobox/MultiCombobox.tsx
vorillaz: Done!

PR #112, opened 2020-08-26
refactor: Replace `css` prop with `sx` prop

3nvi:
### Background

This PR removes emotioon's babel plugin tthat allows  components to specify a `css` prop and replaces it with a similar `sx` prop. The benefits of that are:

-  `css` prop added 1 `<CSSPropInternal>` React element to the component tree of each pounce element, which has now  been  removed. This will improve performance,  since React will have to manage way fewer components

- `css` prop accepts regular CSS & needed  additional configuration to make theming  work,  while `sx` automatically accepts theme-aware props

- `sx` prop has a simple object-like API, where as `css` had a more complex one (see code below)

- Styled-system prop transformations are now handled internally in `sx`, allowing us to delete a lot of custom code

### Changes

- Remove un-needed babel preset
- Globally replaced `css` with `sx` (we  needed  a  different name for overlapping reasons & `sx` is the convention in most  design systems)
- Cleanup un-needed code
- Update all snapshots since the removal of  `css` prop leaves components with 1 emotion class  per component (previously it was 2) cc @vorillaz 

### Testing

- `npm  run test`

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -14,7 +14,6 @@ module.exports = {
     ],
     '@babel/preset-react',
     '@babel/preset-typescript',
-    '@emotion/babel-preset-css-prop',
   ],
   ignore: ['**/*.d.ts'],
   plugins: ['@babel/transform-runtime'],
--- a/package-lock.json
+++ b/package-lock.json
@@ -1533,27 +1533,6 @@
         "stylis": "^4.0.1"
       }
     },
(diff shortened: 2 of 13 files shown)


vorillaz:
@3nvi That's amazing, I love the minimal classes used within the snapshots.
Kudos 🚀

vorillaz (approved):

PR #116, opened 2020-08-26
Feat: DateInput

vorillaz:
### Background

This is the pretty new dead simple `DateInput` component. Eventually, I simplified the API to the bare minimum without using any additional library due to size and limitations. Ditching `react-day-picker` saved us around 9KB minified and lot's of global styles were removed.
Eventually, the DateRange picker component will follow using the same API and the same bits and pieces from the DateInput component.

### Changes

- Remove the `react-day-picker` component.
- Split brand new DateInput component into small individual, composable in a meaningful manner.
- Updated the internal dependencies.
- Created an API that it could get used with the upcoming DateRangeInput component

### Testing

- Visually.
- Through unit tests.

### Screenshots
Screenshot 2020-08-26 at 2 35 43 PM
Screenshot 2020-08-26 at 2 36 01 PM


src/components/DateInput/Day.tsx
```diff
@@ -0,0 +1,70 @@
...
+  &[aria-selected='true'],
+  &[aria-selected='true']:hover {
+    background-color: ${({ theme }) => theme.colors['blue-400']};
+  }
+  &:hover {
+    background-color: ${({ theme }) => theme.colors['navyblue-300']};
+  }
+`;
+
+const DisabledCell = styled(Cell)`
```
3nvi, line 30: with the `<Box>` implementation above, `DisabledCell` doesn't even need to be a component. It could be `<Cell disabled />`.

Styleds for `disabled` get automatically added from `GlobalStyles.tsx` which detects the presence of the selector `&[disabled], &[aria-disabled], &[data-disabled]`

src/components/DateInput/Month.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onDaySelect?: (date: Dayjs) => void;
+}
+
+const Abbr = styled.abbr`
+  font-size: ${({ theme }) => theme.fontSizes.small};
+  font-weight: ${({ theme }) => theme.fontWeights.bold};
+  text-decoration: none;
+`;
+
+const Body = styled.div`
```
3nvi, line 21: all that though `<Box>`. It's easier to reason about (especially during mass updates to the API) when all of our components have the same "base" component.

It's also easier to write them

src/components/DateInput/TimePicker.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <Flex
+      align="center"
+      justify="center"
+      borderTop="1px solid"
+      borderColor="navyblue-300"
+      mx={-6}
+      mt={5}
+    >
+      <Flex align="center" justify="center" p={3} pb={0}>
```
3nvi, line 56: adding `spacing={3}` to t he `Flex` would render the use of the `<Box>` below un-needed

src/utils/GlobalStyles.tsx
```diff
@@ -189,123 +189,6 @@ const GlobalStyles: React.FC = () => {
       -webkit-appearance: none;
     }
 
-    /* ***************** react-day-picker ***************** */
```
3nvi, line 192: nice!

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
+      setOpen(true);
+    },
+    [setOpen]
+  );
+
+  const onDaySelect = useCallback(
+    dateChanged => {
+      const oldDate = dayjs(date);
+      const updated = dateChanged.hour(oldDate.hour()).minute(oldDate.minute());
+      // // @ts-ignore
```
3nvi, line 114: we can potentially remove  the `// @ts-ignore` if you do `const updated = .... as Date`;

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
+    e => {
+      e.preventDefault();
+      setOpen(true);
+    },
+    [setOpen]
+  );
+
+  const onDaySelect = useCallback(
+    dateChanged => {
+      const oldDate = dayjs(date);
```
3nvi, line 112: can we rename to `currentDate`? `oldDate` is a bit misleading

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
-      format={format}
-      value={value}
-      component={(props: TextInputProps) => <TextInput {...props} {...rest} ref={ref} />}
-    />
+    <>
+      <Box position="relative">
+        <Box onClick={onExpand} cursor="pointer">
+          <TextInput
+            {...rest}
+            value={currentDate && dayjs(currentDate).format(format)}
```
3nvi, line 130: when `currentDate` doesn't exist this will be `false`. I think (under the HTML5 spec) it should be an empty string

3nvi: It can't `undefined` or `null` since it would signify a "non-controlled" prop

src/components/DateInput/DateInput.tsx
```diff
@@ -1,57 +1,50 @@
...
-   * A string that represents the formatted Date. It will be parsed according to the given format
-   * and converted to an actual date
+   * A flag that dictates if the component should allow manipulating time
+   */
+  withTime?: boolean;
+
+  /**
+   * A date that that works as a value
    */
-  value: string;
```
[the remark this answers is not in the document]
3nvi: This should actually be a date **string**, since when we "boot up" the component with an existing value , the API will provide a string value.

I think this value should be the default date ISO value, which `dayjs` knows hoow to parse. Thus, when you do:

```ts
  const dateFormatted = date ? dayjs(date) : dayjs();
```

even when `date` is a "string" it will still work.

3nvi (commented):
@vorillaz Amazing work. I can see that it must have been hard to implement and it required a lot of custom implementations. The component works! (I'm always amazed when date picker work 😆 )

I've added some comments  and I'm also letting you know of some UX issues I've found:

- Clicking outside of the DatePicker doesn't hide it automatically (it should IMO)
- There  is a z-index issue (see  attached screenshot)
Screen Shot 2020-08-26 at 4 40 49 PM

- Focus is not locked (I can tab out and go to the background,  that may be minor for now)
- The dates never get focus with "tab" and I can't navigate between  them using arrow keys (The library supports both as shown in this example https://react-day-picker.js.org/docs/getting-started)
- I should be able to "filter" the number for `HH` and `MM` on the dropdown. This helps me since right now I have to "scroll" all the way to where I want (I believe this really simple through the `searchable` prop)

rleighton:
I've been using a version of this testing Indicator Search. The AM/PM selection is VERY hard to use. We use 24 hour timestamps and this requires mental calculations. I also is  very hard to pick items. I very strongly suggest that the AM/PM be replaced with a simple 24 hour picker.

vorillaz:
@3nvi 
>Clicking outside of the DatePicker doesn't hide it automatically (it should IMO)

That's intentional since the input value gets propagated using the Apply / Cancel buttons.

>There is a z-index issue (see attached screenshot)

Fixed.

> Focus is not locked (I can tab out and go to the background, that may be minor for now)

That's somehow harsh to achieve since the input is `readonly`. We can try to enhance the end to end experience in a PR later on.

>The dates never get focus with "tab" and I can't navigate between them using arrow keys (The library supports both as shown in this example https://react-day-picker.js.org/docs/getting-started)

That's the same as the issue mentioned above, that's also something we can do later on as the example works great with the standalone component but not when a "wild" dropdown appears e.g https://react-day-picker.js.org/.

>I should be able to "filter" the number for HH and MM on the dropdown. This helps me since right now I have to "scroll" all the way to where I want (I believe this really simple through the searchable prop)

Done.

@rleighton I agree with you but that's just me. I guess we can stick with that and later on we can add an enhancement with 12h and 24h format. This way we can eventually help our users pick their desired format.

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
vorillaz: That's intentional as well. The thing is that `Date` objects are well structured and play great along with `dayjs`. Since we need to do lots of comparisons, additions, and subtractions to the original value (we need to calculate weeks per month, hours, and minutes and eventually propagate the using the `onChange` handler), using a string values doubles the effort and the complexity.
Furthermore, the API of the component should be agnostic to the format of the value passed through. The consumers are responsible for parsing the string values per case.

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
-) {
+const DateInput: React.FC<DateInputProps> = ({
+  date,
+  format = 'MM/DD/YYYY',
+  withTime,
+  onChangeDate = noop,
+  ...rest
+}) => {
+  const dateFormatted = date ? dayjs(date) : dayjs();
+  const [currentMonth, setCurrentMonth] = useState(dateFormatted);
```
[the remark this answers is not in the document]
vorillaz: Any value passed through the children components is a `dayjs` component cause we need to perform checks and data manipulations. The original date is kept in place cause we need to propagate the final value to the call site.

3nvi:
> That's intentional since the input value gets propagated using the Apply / Cancel buttons.

I actually find this a bit un-expected as a user. The `Modal`, `Dropdown`, `Sidesheet` and `DateInput` (the old one), all closed when the user clicked outside. Is there a particular reason to not have it?

> That's somehow harsh to achieve since the input is readonly. We can try to enhance the end to end experience in a PR later on.

Yeah, there is a `FocusTrap` component that we could use. It's light and would achieve the purpose. What do you think? https://www.npmjs.com/package/react-focus-lock

> That's the same as the issue mentioned above, that's also something we can do later on as the example works great with the standalone component but not when a "wild" dropdown appears e.g https://react-day-picker.js.org/.

Perhaps the above component can help :) I didn't originally realize that you wrote a fully custom picker (that's why I referenced react-day-picker), so yeah we can add this as an enhancement in a separate PR

> @rleighton I agree with you but that's just me. I guess we can stick with that and later on we can add an enhancement with 12h and 24h format. This way we can eventually help our users pick their desired forma

We can also ask @dimitriszoitas about this. If he agrees, we can optionally remove it.

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
+  const [currentDate, setCurrentDate] = useState(date);
+  const [prevDate, setPrevDate] = useState(date);
```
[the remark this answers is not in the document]
vorillaz: Yeap, each time a date is selected is reflected to the input but eventually does not get persisted up until "Apply" button gets clicked. `prevDate` is used in order to restore the previous value upon pressing "Cancel".

vorillaz:
> I actually find this a bit un-expected as a user. The `Modal`, `Dropdown`, `Sidesheet`, and `DateInput` (the old one), all closed when the user clicked outside. Is there a particular reason to not have it?
> 
That's just a different UX approach, nothing special at all. It simplifies things a lot as they're a lot more rather than picking a value for a dropdown. Values are getting reflected/persisted using the action buttons at the dropdown's footer and that's quite brilliant from my perspective. Date pickers and date range pickers are complicated pieces of modules and in the long run, we will have to accommodate a plethora of scenarios. Imagine that the dropdown closes upon clicking outside of the bound area, on top of my head we need to find out solutions for different user actions based on the state of the user's actions up until unexpectedly closing the dropdown.  

`FocusTrap` looks amazing we can give it a try.

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
3nvi: I think using a `React.ref` might be a better call here, since you don't need a re-render when you do `setPrevDate`. It would also simplify react's dependency arrays, but  I might be missing something

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
3nvi: Gotcha, yeah I see that the external API is native only, while the internal uses dayjs. I like it.

src/components/DateInput/DateInput.tsx
3nvi: You resolved it, but I didn't understand why 😅

dimitriszoitas:
> @rleighton I agree with you but that's just me. I guess we can stick with that and later on we can add an enhancement with 12h and 24h format. This way we can eventually help our users pick their desired forma
> 
> We can also ask @dimitriszoitas about this. If he agrees, we can optionally remove it.

I am always in favor of the 24h format but that's because I come from a region that people use the 24h format. I believe people in the US use the AM/PM 12h format and that this would be more convenient for them. If that's not the case, then we can go with a 24h format OR what I would suggest is add a setting (in user settings, or wherever we're going to add the timezone etc) where the user can select the 12 or the 24 format (I think that's what @vorillaz was talking about when he said 'enhancing'.

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
vorillaz: `setPrevDate` does not actually affect rendering as the component get rerendered upon closing the dropdown. It's just a placeholder in order to retrieve the previous value.

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
3nvi: Gotcha, is there a reason then we can't name  the props `value` and `onChange` so they can be automatically handled  by libraries (all oof our inputs have the  same name, that's why)

src/components/DateInput/DateInput.tsx
```diff
@@ -60,21 +53,137 @@ const parseDate = (str: string, format: string): Date | undefined => {
...
+
+            <Flex
+              align="center"
+              justify="center"
+              borderTop="1px solid"
+              borderColor="navyblue-300"
+              mx={-24}
+              mt={20}
+            >
+              <Flex align="center" justify="center" pt={3} spacing={3}>
```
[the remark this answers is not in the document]
vorillaz: Fixed this one according to the new design specs.
https://share.goabstract.com/c434ecdf-65f3-4d3d-bbda-df5758e25f92?mode=build&sha=1bb6d024f756e5ba9d7bc34ea30caf3938b8458a

src/components/DateInput/DateInput.tsx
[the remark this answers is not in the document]
vorillaz: Fixed the API to use props `value` and `onChange`

src/components/DateInput/TimePicker.tsx
vorillaz: Fixed as well.

src/components/DateInput/DateInput.tsx
vorillaz: `// @ts-ignore` was eventually redundant 😓

src/components/DateInput/DateInput.tsx
vorillaz: My bad added an empty string as a default value.

3nvi (approved):

PR #117, opened 2020-08-26
Enhance MultiCombobox UX

3nvi:
### Background

One of the downsides of using MultiCombobox when expecting custom input is that people rarely realize that they need to press "Enter" to save it. In addition, people rarely realize that pressing "Enter" allows them to add more values & when using "pasted" data, that creates problems since their value is lost on blur.

This PR adds UX updates to make their life easier:

- Makes sure to automatically select custom (user entered) values on blur
- Makes sure to automatically tokenize a pasted value containing newlines or commas
- Allows `,`  to be used in the same way as `Enter`

### Changes

- Modify behavior of MultiCombobox
- Make sure to show always render placeholder text
- Add minor accessibility attributes
- Add tests to MultiCombobox

### Testing

- `npm run test`

### Context

We don't automatically convert single-pasted items into a token, since users can sometimes edit those. When pasting multiple values, users rarely want to "change" a particular one, so we automatically convert them to tokens. Agreed upon with our designer.

### DIFF
--- a/jest/utils/render.tsx
+++ b/jest/utils/render.tsx
@@ -2,6 +2,6 @@ import React from 'react';
 import ThemeProvider from '../../src/utils/ThemeProvider';
 import { render as rtlRender, RenderOptions as RtlRenderOptions } from '@testing-library/react';
 
-export const renderWithTheme = (element: React.ReactElement, options: RtlRenderOptions) => {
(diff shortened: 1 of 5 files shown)


alexmylonas (approved):
Nice job adding test for new case as well!

PR #119, opened 2020-09-01
feat: add iconProps for TextInput

vorillaz:
### Background

Drilling down the icon props used for styling.

### Testing

- Manually and visullay
- `npm run test`

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -5,6 +5,13 @@ import { InputControl, InputElement, InputLabel } from '../utils/Input';
 import Icon, { IconProps } from '../Icon';
 import Flex from '../Flex';
 
+export type TextIconProps = {
+  color?: IconProps['color'];
+
+  /** The size of the icon. Can be 18px or 24px */
+  size?: IconProps['size'];
+};
+
 export type TextInputProps = NativeAttributes<'input'> & {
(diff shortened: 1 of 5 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -20,10 +20,13 @@ export type TextInputProps = NativeAttributes<'input'> & {
 
   /** The icon present on the input  */
   icon?: IconProps['type'];
+
+  /** Additional icon props  */
+  iconProps?: Omit<IconProps, 'type' | 'ref'>;
```
3nvi, line 25: Hmm I feel that `IconProps` is a non-restricting enough. Currently  I can add `iconProps={{ position: 'fixed' }}` and totally mess with your layout  😄 

Should we perhaps either allow a fixed set of icon props or split them into separate ones `i.e. `iconSize`, `iconColor` instead of a single `iconProps`)

vorillaz: And it's done!

3nvi (approved):
I'm approving since the problem has been addressed.

 I'd generally  love to see a  similar thing happening in `TextArea`  component as  well

vorillaz:
> I'm approving since the problem has been addressed.
> 
> I'd generally love to see a similar thing happening in `TextArea` component as well

Will do soon :)

PR #120, opened 2020-09-03
fix: make sure collapse animation doesn't hinder clicking

3nvi:
### Background

Currently when something is animating, the items "behind" the animation are not clickable until the animation finishes. In cases where the item gets removed from the screen, you see it disappearing but UI behind it is still not clickable until the animation finishes.

This gets fixed by  adding `pointer-events: none` to the item that's "disappearing"

### Changes

- Instantly add `pointer-events: none` to the animating item

### Testing

- Testing steps


alexmylonas (approved):
LGTM!

vorillaz (approved):

PR #123, opened 2020-09-10
Fix: daysjs parsing dates for DateRangeInput

vorillaz:
### Background
When value is empty there was a weird highlighting issue with the selected dates. This one parses just the values provided without fallbacks for new values.

### Testing
- Visually and manually
- `npm run test`

### DIFF
--- a/src/components/DateRangeInput/DateRangeInput.tsx
+++ b/src/components/DateRangeInput/DateRangeInput.tsx
@@ -68,12 +68,16 @@ export interface DateRangeInputProps {
   onChange: (date: Date[]) => void;
 }
 
-const convertToDayjs = (value: Date[]): [Dayjs, Dayjs] => {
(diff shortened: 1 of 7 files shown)


src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -68,12 +68,16 @@ export interface DateRangeInputProps {
   onChange: (date: Date[]) => void;
 }
 
-const convertToDayjs = (value: Date[]): [Dayjs, Dayjs] => {
+const dateToDayjs = (value: Date[]): Dayjs[] => {
```
3nvi, line 71: I feel you don't "need" all these typings. TS is smart enough 😆

3nvi (approved):
LGTM 🌮

PR #125, opened 2020-09-14
fix: Add bulk-upload and source-code icons

vorillaz:
### Background

Added two teeny tiny icons used for the Bulk uploader redesign


alexmylonas (approved):
LGTM!

PR #128, opened 2020-09-16
feat: allow custom typing extension to support custom icons

3nvi:
### Background

Currently, pounce allows  you  to have custom icon definitions, by overriding the `icons` prop in the `theme` that gets passed  to  the `ThemeProvider`. 

So far though, you'd always be getting an invalid TS error, stating that this icon doesn't exist. To fix that, pounce now exposes a `CustomIcons` type which you can override locally in your project (using a local declaration file) in order to extend the set of known `icon`  keys that Pounce is aware of.

### Changes

- Make sure all icon keys are taken from `Theme`
- Export an empty `CustomIcons` interface in order to allow overrides
- Update docs

### Testing

- Installed it in a project and managed to get it working  on external projects by following the README instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,11 +122,54 @@ export const myTheme = {
 };
 ```
 
+By default, pounce ships with dashboard-related icons, but you may add new ones by extending the
+`icons` keyy of the theme. For example, if you want to add a new icon named `my-icon` you would
+override the theme and pass it again to the `<ThemeProvider>` component.
+
+```js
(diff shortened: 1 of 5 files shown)


vorillaz:
Hey, that's super cool. On our side, though I don't think we shall consume the custom icons in Panther, we are going to lose the source of truth as we are going to split the actual icon set in two places.
Btw take a look at the test suite 😞

vorillaz (approved):

PR #129, opened 2020-09-16
Fix global date mocks for time related specs

vorillaz:
### Background
Since presets are dynamically generated using the current date snapshots are failing, thus we need to mock the global `Date` constructor. More info about the `dayjs` implementation can be found [here](https://github.com/iamkun/dayjs/issues/500)

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "jest-emotion": "^10.0.32",
     "jest-extended": "^0.11.5",
     "lint-staged": "^10.1.0",
+    "mockdate": "^3.0.2",
     "prettier": "^2.0.5",
     "react": "^16.11.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "fuzzaldrin": "^2.1.0",
     "jest-axe": "^3.5.0",
     "lodash": "^4.17.19",
+    "mockdate": "^3.0.2",
```
3nvi, line 48: can we  add it as a dev dependency?

3nvi (approved):
🎉 LGTM but please add it under `devDependencies`.

package.json
vorillaz: Oops.

PR #130, opened 2020-09-16
Feat subtract icon

3nvi:
### Background

Adds a new `subtract` icon  that compliments the existing `add` icon

### Changes

- Add substract/minus icon
- Add self-closing tag to `<path>` HTML elements


alexmylonas (approved):
LGTM!

PR #134, opened 2020-09-18
Fix datewrapper enhancements

vorillaz:
### Background

This one adds a shadow for the date wrapper component used from `DatePicker` and `DateRange` picker components/. Also fixes some placement issues due to the absolute positioning of the wrapper. 
`reachui/popover` gets bumped as it was explicitly missing from the dependencies.


src/utils/GlobalStyles.tsx
```diff
@@ -209,6 +209,9 @@ const GlobalStyles: React.FC = () => {
       --reach-tabs: 1;
     }
 
+    [data-reach-popover] {
```
3nvi, line 212: can we not add this global style, but  instead add a `style= {{ ... }}` prop in the associated Popover uses? 

For example, you could add it as part of `transitions` in the `enter` key. This recommendation is purely to avoid un-expectefd side-effects

3nvi (approved):
Looks good @vorillaz . 1  minor comment 🎉

src/utils/GlobalStyles.tsx
vorillaz: @3nvi That was my initial thought in order to avoid leaking styles, but the actual `reach-popover` is injected inside the body and that's why we have additional `data-reach` selectors inside `GlobalStyles.tsx`
eg 
https://github.com/panther-labs/pounce/blob/2fb096fceead34592c8ad598ca13c7c8d4191edf/src/utils/GlobalStyles.tsx#L215-L223

Any thoughts on that?

3nvi: if it's already like that then fine. I just wanna make sure that I limit the blast radius of the changes

3nvi: Let's verify that tooltips overlap correctly and that or comboboxes inside modals DateInputs inside Modals work ok (z-index wise). If that's fine, then we are good.

PR #136, opened 2020-09-26
fix: change `X` icon to `!` on the alert "error"

3nvi:
### Background

Lot of people have complained that  our "X" icon looks like a button and they attempt to click it. With this in mind, the alert error style, will use an "!" icon  that will not confused users.

### Changes

- Replace "X" icon with "!"

### Testing

- N/A


alexmylonas (approved):
LGTM! 👍

PR #145, opened 2020-10-21
Feat: Popover component

3nvi:
### Background

This PR introduced a `Popover` component, which is part of a compound component family, comprising of:

- `Popover`: a wrapper (like `Dropdown`) that manages the state and decides what's shown and what's not
- `PopoverTrigger`: a trigger element (like `DropdownButton`) which acts as the trigger of the popover
 `PopoverContent`: the actual popover that gets rendered

### Features

- Allow `Popover` to render as initially visible
- Allow `PopoverMenu` to have custom positioning
- Allow `offset` distance between the trigger and the popover
- Toggle popover on trigger click
- Toggle popover on trigger <Enter> keypress
- Close popover on click outside
- Close popover on ECS key press
- Move focus to popover as soon as its opened
- Accept render props components, exposing its internal state

### Changes

- Adds Popover compound component family

### Testing

- Snapshots
- TBA

### DIFF
--- a/src/components/Popover/Popover.test.tsx
+++ b/src/components/Popover/Popover.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { renderWithTheme } from 'test-utils';
 import Popover from './Popover';
 import PopoverTrigger from './PopoverTrigger';
-import PopoverMenu from './PopoverMenu';
+import PopoverMenu from './PopoverContent';
 import Button from '../Button';
 
 describe('Popover', () => {
--- a/src/components/Popover/Popover.test.tsx
+++ b/src/components/Popover/Popover.test.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { renderWithTheme } from 'test-utils';
+import Popover from './Popover';
+import PopoverTrigger from './PopoverTrigger';
+import PopoverMenu from './PopoverMenu';
+import Button from '../Button';
+
+describe('Popover', () => {
+  it('matches snapshot', () => {
+    const { container } = renderWithTheme(
+      <Popover>
+        <PopoverTrigger as={Button}>Click me</PopoverTrigger>
+        <PopoverMenu>Boom! I am The popup</PopoverMenu>
+      </Popover>
+    );
+
+    expect(container).toMatchSnapshot();
+  });
+});
--- a/src/components/Popover/Popover.test.tsx
+++ b/src/components/Popover/Popover.test.tsx
@@ -1,19 +1,184 @@
 import React from 'react';
-import { renderWithTheme } from 'test-utils';
+import {
+  fireClickAndMouseEvents,
+  fireEvent,
+  renderWithTheme,
+  waitForElementToBeRemoved,
+  waitMs,
+} from 'test-utils';
 import Popover from './Popover';
 import PopoverTrigger from './PopoverTrigger';
 import PopoverMenu from './PopoverContent';
 import Button from '../Button';
+import Text from '../Text';
 
 describe('Popover', () => {
-  it('matches snapshot', () => {
+  it('matches snapshot when closed', () => {
     const { container } = renderWithTheme(
-      <Popover>
+      <Popover id="test">
         <PopoverTrigger as={Button}>Click me</PopoverTrigger>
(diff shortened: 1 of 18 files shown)


src/components/Popover/usePopoverContentAlignment.tsx
```diff
@@ -0,0 +1,136 @@
+import { PopoverContentProps } from './PopoverContent';
+import { Position, getCollisions } from '@reach/popover';
+
+export const positionBottomLeft = (offset: [number, number]): Position => (
```
vorillaz, line 4: That's odd and overengineered IMHO.
It also means that we should potentially detect window size changes and rearrange the popover accordingly.
Ideally, I would expect to add a "placeholder element" underneath the popover triggering element and use absolute positioning for the content. I am not also sure if we would actually use the placement options for now.

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,125 @@
...
+   * @default left
+   */
+  alignment?: 'bottom-left' | 'bottom-center' | 'bottom-right' | 'left-center' | 'right-center';
+
+  /**
+   * The spacing between the trigger and the popover in px, for the horizontal & vertical axis
+   *
+   * @default [0, 12]
+   */
+  offset?: [number, number];
```
vorillaz, line 25: I would allow offsetting tbh, I would suggest strictly following the designs.

src/components/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { renderWithTheme } from 'test-utils';
+import Popover from './Popover';
+import PopoverTrigger from './PopoverTrigger';
+import PopoverMenu from './PopoverMenu';
+import Button from '../Button';
+
+describe('Popover', () => {
+  it('matches snapshot', () => {
+    const { container } = renderWithTheme(
+      <Popover>
```
vorillaz, line 11: Ι guess we should snapshot the Popover when it's open as well.

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isOpen?: boolean;
+
+  /**
+   * The id of the popover. Helps setting up aria controls between popover content & trigger
+   */
+  popoverId?: string;
+}
+
+interface PopoverContext {
+  popoverId?: string;
```
vorillaz, line 17: Semantically the `popoverId` is vital so we can use [auto-id](https://reach.tech/auto-id) as a default value.

src/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,125 @@
...
+    { alignment = 'bottom-left', offset = defaultOffset, children, ...rest },
+    forwardedRef
+  ) {
+    const popoverAlignment = usePopoverContentAlignment({ alignment, offset });
+    const { popoverId, isOpen, triggerRef, popoverRef, close } = usePopoverContext();
+
+    // Close on clicks outside. We only want to close if a click occurred outside the popover
+    // and its trigger. We also add `capture` events to avoid some race conditions on window-attached
+    // events
+    React.useEffect(() => {
```
vorillaz, line 44: I guess we can move this part to a hook in order to control and reuse the behavior.

vorillaz (commented):
Overall looks neat, with a minor few comments here and there.

src/components/Popover/Popover.tsx
3nvi: I've removed custom overriding since it's 1 less thing to worry about 😅

src/components/Popover/PopoverContent.tsx
3nvi: Implemented

src/components/Popover/usePopoverContentAlignment.tsx
vorillaz: Gotcha, I guess that's the way to go in order to avoid frustration and CSS overrides.

src/components/Popover/Popover.tsx
vorillaz: 👍

vorillaz (approved):
Looking neat, just drop me a line when the version gets bumped for Panther.

PR #146, opened 2020-10-23
feat: remove `offset` and add more `alignment` options to Popover

3nvi:
### Background

After reviewing the `offset` prop of Popover, I came to the realization that it's a bit weird to have `alignment` and `offset` since these are conflicting props. 

What made sense to me was to only have `alignment` props and then allow the user to specify a spacing through a `margin` prop on the customized content that he/she adds to the `Popover` (see example in docs)

To make sure that the removal of `offset` doesn't cause issues, I increased the number of alignments so that all possible scenarios could be handled. Namely:

    'bottom-left'
    'bottom-center'
    'bottom-right'
    'top-left'
    'top-center'
    'top-right'
    'left-bottom'
    'left-center'
    'left-top'
    'right-bottom'
    'right-center'
    'right-top'

This PR includes just that

### Changes

- Remove `offset` and allow the user to add `margin` to their own component
- Add more `alignment` options
- Cleanup duplications in alignments

### Testing

- Existing tests + new snapshots


src/components/Popover/Popover.test.tsx
```diff
@@ -35,17 +35,6 @@ describe('Popover', () => {
...
-  it('matches snapshot with offset', () => {
-    const { container } = renderWithTheme(
-      <Popover isOpen id="test">
-        <PopoverTrigger as={Button}>Click me</PopoverTrigger>
-        <PopoverMenu offset={[10, 10]}>Boom! I am the popup</PopoverMenu>
-      </Popover>
-    );
-
-    expect(container).toMatchSnapshot();
-  });
```
alexmylonas, line 47: We cant have any snapshot test with this change?

alexmylonas (approved):

vorillaz (approved):

src/components/Popover/Popover.test.tsx
3nvi: we have plenty  more :D

PR #148, opened 2020-11-02
fix: stop loading the entire lodash library for a single method

3nvi:
### Background

Currently we bundle all of lodash, even though we use a single method of it. This increases our bundle size by at least 50KB!

### Changes

- Change `import { X } from 'lodash'` to `import X from 'lodash/X'`

### Testing

- `npm run test`


vorillaz (approved):
Mea culpa.

PR #152, opened 2020-11-09
Fix raising input labels

vorillaz:
### Background

Input labels were a bit off, raised upon values that made no sense. This PR introduces a tiny helper to accommodate empty input values (`null`, `undefined`, `''`). Even a single space can be valid though.

### Changes

- Added `isEmpty` helper
- Consumed the helper for `TextInput` and `NumberInput` components.
- Added helper tests and examples.

### Testing

- Visually
- `npm run test`

### DIFF
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -72,6 +72,18 @@ export const isBrowser = typeof window !== 'undefined' && typeof window.document
 
 export const noop = (): void => {};
 
+export const isEmpty = (value?: string | number | readonly string[] | undefined): boolean => {
+  if (value === null || value === undefined) {
+    return true;
+  }
+
+  if (typeof value === 'string' && value === '') {
+    return true;
+  }
+
+  return false;
+};
+
 /**
  * A function that generates a preset with dynamic dates.
  * @returns {Object} An object with Dayjs presets
--- a/src/utils/helpers.ts
+++ b/src/utils/helpers.ts
@@ -72,12 +72,22 @@ export const isBrowser = typeof window !== 'undefined' && typeof window.document
 
 export const noop = (): void => {};
 
-export const isEmpty = (value?: string | number | readonly string[] | undefined): boolean => {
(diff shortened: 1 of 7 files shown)


src/utils/helpers.ts
```diff
@@ -72,6 +72,18 @@ export const isBrowser = typeof window !== 'undefined' && typeof window.document
...
 
+export const isEmpty = (value?: string | number | readonly string[] | undefined): boolean => {
+  if (value === null || value === undefined) {
+    return true;
+  }
+
+  if (typeof value === 'string' && value === '') {
+    return true;
+  }
+
```
3nvi, line 83: can we also add:

```ts
if (Array.isArray(value) && value.length === 0) {
  return true
}

if (typeof value === 'object' && Object.keys(value).length === 0) {
  return true
}
```

and update tests accordingly?

src/utils/helpers.ts
```diff
@@ -72,6 +72,18 @@ export const isBrowser = typeof window !== 'undefined' && typeof window.document
 
 export const noop = (): void => {};
 
+export const isEmpty = (value?: string | number | readonly string[] | undefined): boolean => {
```
3nvi, line 75: to avoid lodash clashes (and possibly developer mistakes), can we rename this to `isEmptyValue`?

If we do that, perhaps the comment below about arrays and objects might not be needed

3nvi (approved):
LGTM 🎉 

Couple of minor changes that I'd recommend for clarity, but nothing too big. I'm pre-approving cause I don't feel strongly about them :)

src/utils/helpers.ts
vorillaz: Yeap, nice catch. Updated.

src/utils/helpers.ts
vorillaz: That was my initial thought, your suggestion makes totally sense.
Fixed!

PR #153, opened 2020-11-11
feat: added AbstractButton that provides the ability to clear all items

alexmylonas:
### Background

It was requested to have the option to clear all items in `MultiComboBox`. I added this option to be display only when at least 6 items are added to the combobox.

### Changes

- Added an AbstractButton displayed conditionally
- Added tests for functionality

Screenshot 2020-11-11 at 10 35 13 PM

### Testing

- Locally & visually

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -11,7 +11,6 @@ import Tag from './Tag';
 import { typedMemo } from '../../utils/helpers';
 import Menu from '../utils/Menu';
 import AbstractButton from '../AbstractButton';
-import Text from '../Text';
 
 export type MultiComboboxProps<T> = {
   /** Callback when the selection changes */
@@ -87,6 +86,11 @@ export type MultiComboboxProps<T> = {
    * item will be added to the selection. If not, then this item won't be added
    * */
   validateAddition?: (userEnteredInput: string, existing: T[]) => boolean;
+
+  /**
+   *
+   */
+  canClearAllAfter?: number;
 };
 
 const stateReducer = (state: DownshiftState<any>, changes: StateChangeOptions<any>) => {
@@ -132,6 +136,7 @@ function MultiCombobox<Item>({
   validateAddition = () => true,
   maxHeight = 300,
   maxResults,
+  canClearAllAfter = 5,
   invalid,
   hidden,
   ...rest
@@ -155,7 +160,7 @@ function MultiCombobox<Item>({
     }
   };
 
-  const clearItems = () => {
+  const clearSelectedItems = () => {
     onChange([]);
   };
 
@@ -304,7 +309,7 @@ function MultiCombobox<Item>({
                     ))}
                   </Flex>
                 )}
-                {value.length > 5 && (
+                {canClearAllAfter && value.length >= canClearAllAfter && (
                   <AbstractButton
                     color="teal-300"
                     zIndex={2}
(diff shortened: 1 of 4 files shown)


src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -153,6 +155,10 @@ function MultiCombobox<Item>({
     }
   };
 
+  const clearItems = () => {
```
3nvi, line 158:
```suggestion
  const clearSelectedItems = () => {
```

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -298,6 +304,22 @@ function MultiCombobox<Item>({
                     ))}
                   </Flex>
                 )}
+                {value.length > 5 && (
```
3nvi, line 307: Instead of arbitrarily choosing `5`, can we control the presence of this button through a prop?

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -298,6 +304,22 @@ function MultiCombobox<Item>({
...
+                  <AbstractButton
+                    color="teal-300"
+                    zIndex={2}
+                    position="absolute"
+                    bottom={0}
+                    right={18}
+                    mb={1}
+                    onClick={clearItems}
+                  >
+                    <Text fontSize="small" fontStyle="italic" textDecoration="underline">
```
3nvi, line 317: We can skip the `Text` here since we can have all these props directly on `AbstractButton`

3nvi (commented):
Nice work @alexmylonas . Let's control this with a prop and update/write a snapshot test as well

src/components/MultiCombobox/MultiCombobox.tsx
alexmylonas: I think I would like this to be on by default and have some default length.

But i will add something like `canClearAllAfter: number`

src/components/MultiCombobox/MultiCombobox.tsx
alexmylonas: Done

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -298,6 +309,23 @@ function MultiCombobox<Item>({
                     ))}
                   </Flex>
                 )}
+                {canClearAllAfter && value.length >= canClearAllAfter && (
```
3nvi, line 312: `canClearAllAfter` will always be defined

```suggestion
                {value.length >= canClearAllAfter && (
```

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -85,6 +86,11 @@ export type MultiComboboxProps<T> = {
    * item will be added to the selection. If not, then this item won't be added
    * */
   validateAddition?: (userEnteredInput: string, existing: T[]) => boolean;
+
+  /**
+   *
```
3nvi, line 91: I'm not feeling strongly about the length thing. For me I'd rather we either render it or we don't. I don't feel strongly though, so I'll defer to whatever @vorillaz believes

vorillaz: Same here, I would suggest passing this one as `prop`. Something like `clearAfter?: number`. When it's undefined we shall not show "clear all" option trigger. IMHO it makes sense from a UX perspective.
We want to clear all the options for Log type filtering, but we don't need this one when creating a rule as the action is more regulated, it's a singleton and we are forcing the user to think twice about the input.
On the other hand, it makes more sense to use this one for quick actions or when the UI is not so "responsive" as the filters mentioned above.

vorillaz (approved):
LGTM, just a minor note to check this one how it looks with solid and outlined input elements.

alexmylonas:
Thanks for your feedback guys! 

I make the option optional and when not provided, the Clear all option won't be rendered

Check screenshot when is set: 
Screenshot 2020-11-12 at 4 58 04 PM

PR #155, opened 2020-11-12
fix: add hours range from 00 to 23 in 24h format

vorillaz:
### Background

This one adds a new range for 24h format date pickers.

### Screenshot
Screenshot 2020-11-12 at 9 20 37 PM

### Changes

- For 24h format we are adding hours as 00 ... 23

### Testing

- `npm run test`


3nvi (approved):
Looks good @vorillaz

PR #159, opened 2020-12-01
Fix: DateRangePicker scroll overflow

3nvi:
### Background

Sometimes, when the `DateInput` or `DateRangeInput` is rendered close to the bottom of the screen, the picker overflows and no matter how much you scroll you can't reach it.

This is illustrated by the screenshot below. 

Screen Shot 2020-12-01 at 4 49 51 PM

This PR fixes it, by removing `absolute` styles and opting for a different solution.

Closes https://github.com/panther-labs/panther-enterprise/issues/1562

### Changes

- Fix an issue with `Flex` where when both `spacing` and `sx` where specified, only one of them would apply
- Remove styling from `DateWrapper` and add a custom `maxWidth` on the `TimePicker`'s inputs, so they don't stretch to 195px (default for Chrome)
- Move `Flex` wrapper to `TimePicker` so that we don't need to duplicate code for both `DateInput` and `DateRangeInput`

### Screenshots

Screen Shot 2020-12-01 at 4 52 50 PM

and 

Screen Shot 2020-12-01 at 4 53 18 PM

### Testing

- `npm run test`


vorillaz (approved):
Nicely done.

PR #162, opened 2021-01-15
fix: removed maxWidth from Modal

alexmylonas:
### Background

We need to have a Modal that exceeds the existing `maxWidth`

### Changes

- Remove `maxWidth` from Modal


3nvi (approved):
LGTM

PR #165, opened 2021-01-26
Add timezone support to DatePickers

kouknick:
## Description
This PR adds timezone support and fixes issues related to the Date picker components.
### Changes 
- Add timezone prop to Datepicker and DateRange picker
- Fix invalid time selected in time pickers when 12H format is enabled
- Avoid resetting time when the Date changes
- Do not allow more than two dates to be passed as props to DateRangePicker
- Round the selected Date in DateInput to avoid inconsistency

### Checklist:

- [x] Make sure you give an high-level overview of the problem & the solution
- [x] Make sure the related issue is referenced
- [x] Make sure screenshots are added if the UI is changed in any way
- [x] Make sure all of the significant logic is covered by tests
- [x] Make sure your changes are accessible (addition of aria-attributes & correct HTML semantics)
- [x] Make sure you reference the related design document if it exists
- [ ] Make sure telemetry is added (analytics & errors)


vorillaz (approved):
Overall looking amazing, thanks a lot for addressing these issues.

PR #173, opened 2021-03-01
feat: inline MultiComboBox input field

kouknick:
## References

closes https://app.asana.com/0/1199975676341842/1199936728416968

## Background

This PR changes the way we render the `TextInput` in multi select `Dropdowns`, in order to keep the `TextInput` in the same line as the `Tags` and only move to a separate line when the user's input can't fit in the existing line anymore.

## Screencast

https://user-images.githubusercontent.com/15084305/109520818-4b8af800-7ab5-11eb-95dc-96144576c536.mov

## Testing

- Locally

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -314,7 +314,7 @@ function MultiCombobox<Item>({
                       minWidth="60px"
                       position={isOpen && searchable ? 'relative' : 'initial'}
                     >
-                      {isOpen && (
+                      {isOpen && searchable && (
                         <InputElement
                           as="span"
                           px={2}
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -224,7 +224,7 @@ function MultiCombobox<Item>({
             placeholder: !value.length ? placeholder : '',
           }),
           ...(searchable && {
-            placeholder: hideLabel && value.length && !isOpen ? '' : placeholder,
+            placeholder: (hideLabel && value.length && !isOpen) || value.length ? '' : placeholder,
             p: isOpen ? 1 : null,
           }),
           position: 'absolute',
@@ -311,7 +311,6 @@ function MultiCombobox<Item>({
                       as="li"
                       maxWidth="100%"
                       flexGrow={1}
-                      minWidth="60px"
                       position={isOpen && searchable ? 'relative' : 'initial'}
                     >
                       {isOpen && searchable && (
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -222,13 +222,12 @@ function MultiCombobox<Item>({
             onMouseDown: toggleMenu,
             readOnly: true,
             placeholder: !value.length ? placeholder : '',
-            position: !value.length ? 'static' : 'absolute',
           }),
           ...(searchable && {
-            placeholder,
-            mt: (isOpen && value.length) || (isOpen && value.length && allowAdditions) ? -4 : 0,
-            position: isOpen || (isOpen && value.length && allowAdditions) ? 'static' : 'absolute',
+            placeholder: hideLabel && value.length && !isOpen ? '' : placeholder,
+            p: isOpen ? 1 : null,
           }),
+          position: 'absolute',
           width: '100%',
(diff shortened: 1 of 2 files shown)


src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -308,8 +307,33 @@ function MultiCombobox<Item>({
...
+                          as="span"
+                          px={2}
+                          py={0}
+                          standalone={hideLabel}
+                          visibility="hidden"
+                          whiteSpace="pre"
+                        >
+                          {inputValue}
+                        </InputElement>
+                      )}
```
kouknick, line 328: This is the hack that will force the input to move to the next line when the content doesn't fit in the first line.
I used a hidden span that has the same content as the input in order to track the width of the input.

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -222,13 +222,12 @@ function MultiCombobox<Item>({
...
-            position: !value.length ? 'static' : 'absolute',
           }),
           ...(searchable && {
-            placeholder,
-            mt: (isOpen && value.length) || (isOpen && value.length && allowAdditions) ? -4 : 0,
-            position: isOpen || (isOpen && value.length && allowAdditions) ? 'static' : 'absolute',
+            placeholder: hideLabel && value.length && !isOpen ? '' : placeholder,
+            p: isOpen ? 1 : null,
           }),
+          position: 'absolute',
```
kouknick, line 230: Changed position to absolute:
- When the dropdown is closed the input will cover the entire dropdown.
- When the dropdown is opened it will be inlined in the `Tag` list and have the same width and height as the hidden span.

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -222,13 +222,12 @@ function MultiCombobox<Item>({
             onMouseDown: toggleMenu,
             readOnly: true,
             placeholder: !value.length ? placeholder : '',
-            position: !value.length ? 'static' : 'absolute',
           }),
           ...(searchable && {
-            placeholder,
-            mt: (isOpen && value.length) || (isOpen && value.length && allowAdditions) ? -4 : 0,
-            position: isOpen || (isOpen && value.length && allowAdditions) ? 'static' : 'absolute',
+            placeholder: hideLabel && value.length && !isOpen ? '' : placeholder,
```
kouknick, line 227: This fixes a visual bug when the label was hidden the input placeholder was rendered behind the Tags.
image

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -222,13 +222,12 @@ function MultiCombobox<Item>({
...
+            p: isOpen ? 1 : null,
```
3nvi, line 228: why are we adding a padding here? Is it so we "align" the input to the blue tags?

src/components/MultiCombobox/MultiCombobox.tsx
3nvi, line 227: is it easy to ellipsify the placeholder? I acn see the placeholder text being cut off sometimes.

See [here](https://codepen.io/bradleytaunt/pen/OgpzyY/) for an example. We can also just remove the placeholder when **at least one value** has been added (i.e. only show it when no values are present)

src/components/MultiCombobox/MultiCombobox.tsx
3nvi: Is there a way to have the input be present in the same line in some edge cases?

Screen Shot 2021-03-02 at 12 51 04 PM
Screen Shot 2021-03-02 at 12 51 06 PM

I would have expected the following:

- The input shows in the same line for the first 1-2 characters
- It moves in its own line when it doesn't fit anymore

3nvi (commented):
Great work @kouknick 🌮 Some small comments/questions

src/components/MultiCombobox/MultiCombobox.tsx
kouknick: Correct.

src/components/MultiCombobox/MultiCombobox.tsx
kouknick: This is because i added `minWidth="60px"` to the input. I'll remove it.

3nvi (approved):
Amazing stuff @kouknick . Thanks a lot 🎉 🚢

PR #186, opened 2021-03-19
Prefix for TextInput

vorillaz:
### Background

Add support for prefix within the text input components.

https://user-images.githubusercontent.com/1022166/111817114-48ca3880-88e6-11eb-8867-08aa50c5cbc1.mov

### Changes

- Add support and fix the Text Input component 
- Add and fix snapshot testing.

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { NativeAttributes } from '../Box';
+import Box, { NativeAttributes } from '../Box';
 import { slugify, isEmptyValue } from '../../utils/helpers';
 import { InputControl, InputElement, InputLabel } from '../utils/Input';
 import Icon, { IconProps } from '../Icon';
@@ -15,6 +15,9 @@ export type TextIconProps = {
 export type TextInputProps = NativeAttributes<'input'> & {
   /** The label that is associated with this input */
   label: string;
+
+  /** The decorating prefix used for  decorating issues*/
+  prefix?: string;
 
   /** The variant of the input that decides the colors */
   variant?: 'solid' | 'outline';
@@ -45,6 +48,7 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(function Te
     required,
     disabled,
     id,
+    prefix,
     hidden,
     name,
     icon,
@@ -57,6 +61,7 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(function Te
   ref
 ) {
   const identifier = id || name || slugify(label);
+  const isEmpty = isEmptyValue(value);
   return (
     <InputControl
       variant={variant}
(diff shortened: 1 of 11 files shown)


src/components/utils/Input/InputControl.tsx
```diff
@@ -61,6 +61,10 @@ const InputControl: React.FC<InputControlProps> = ({
       }}
       _focusWithin={{
         borderColor: !invalid ? 'blue-400' : undefined,
+        '[data-prefix]': {
```
3nvi, line 64: can you not use  `span` or `& > span` here to avoid the `data-prefix` attribute?

src/components/TextInput/TextInput.tsx
```diff
@@ -66,22 +71,41 @@ const TextInput = React.forwardRef<HTMLInputElement, TextInputProps>(function Te
...
+            data-prefix
+            fontSize="medium"
+            fontWeight="medium"
+            as="span"
+            height="100%"
+            color="navyblue-100"
+            pt={5}
+            pl={4}
+            pb={2}
+            position="relative"
```
3nvi, line 85: do we need relative positioning here?

src/components/utils/Input/InputElement.tsx
```diff
@@ -24,7 +24,6 @@ const InputElement = React.forwardRef<HTMLInputElement | HTMLTextAreaElement, In
       <Box
         ref={ref}
         as="input"
-        verticalAlign="top"
```
3nvi, line 27: I believe the reason you had to remove this is because your prefix is a `span`. If it was a `div` or anything else you wouldn't need to.

I'll have to verify that this hasn't caused problems in other places

3nvi (commented):
Good work on this @vorillaz 🌮 . Will do a functional testing and update my review with my findings.

src/components/utils/Input/InputElement.tsx
vorillaz: The cleanup was due to Firefox,  validating the input. `verticalAlign` applies to `inline-block` elements and the `input` elements are just `inline`.

src/components/TextInput/TextInput.tsx
```diff
@@ -16,6 +16,9 @@ export type TextInputProps = NativeAttributes<'input'> & {
   /** The label that is associated with this input */
   label: string;
 
+  /** The decorating prefix used for  decorating issues*/
+  prefix?: string;
```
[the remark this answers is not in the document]
vorillaz: Hey @3nvi , thanks for the feedback. I thought so during the implementation but eventually that's more a consideration for the consumer.
In our case eventually the `formik field component` should perform any value manipulation.
What do you think about that?

src/components/TextInput/TextInput.tsx
vorillaz: Fixed!

src/components/utils/Input/InputControl.tsx
vorillaz: Done!

src/components/TextInput/TextInput.tsx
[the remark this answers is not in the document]
3nvi: Yeah I agree. Let's proceed as is

src/components/utils/Input/InputElement.tsx
3nvi: Gotcha. Ok, resolving this one and will re-open if I find that something looks misaligned :)

3nvi (approved):
Tested it on all major browsers. Looks good 🎉

AngelGzS (approved):

PR #204, opened 2021-04-14
Datepicker clear

vorillaz:
### Background

This PR introduces "a clear button" for DateInput and DateRangeInput components. When pressed the values are set to `undefined` and `[undefined, undefined]`, `onChange` callback is triggered on "Apply"

### Changes

- Add TextLink component
- Add clear option for the components
- Fix and enhance specs

### Screenshots

https://user-images.githubusercontent.com/1022166/114758949-39fc6780-9d66-11eb-9c5a-1b0e642665c3.mov

https://user-images.githubusercontent.com/1022166/114758957-3c5ec180-9d66-11eb-8ea3-4260d68018c2.mov


src/components/DateInput/DateInput.tsx
```diff
@@ -102,16 +112,20 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
     close();
   }, [close, value, timezone, setCurrentDate]);
 
+  const onClear = useCallback(() => setCurrentDate(undefined), [setCurrentDate]);
```
3nvi, line 115: To follow the convention above, I'd do 

```suggestion
  const onClear = useCallback(() => setCurrentDate(dateToDayjs(undefined)), [setCurrentDate]);
```

src/components/DateInput/DateInput.tsx
```diff
@@ -102,16 +112,20 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
...
   const onApply = useCallback(
     e => {
-      if (!currentDate) {
+      if (!currentDate && disableReset) {
         return;
       }
       e.preventDefault();
-      onChange(currentDate.startOf(withTime ? 'minute' : 'day').toDate());
+      onChange(
+        !currentDate ? currentDate : currentDate.startOf(withTime ? 'minute' : 'day').toDate()
```
3nvi, line 124: out ouf curiosity, since `currentDate` is undefined, wouldn't that work? The ternary would quickly yield `undefined`, no?

```suggestion
        currentDate?.startOf(withTime ? 'minute' : 'day').toDate()
```

src/components/DateInput/DateInput.tsx
```diff
@@ -102,16 +112,20 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
     close();
   }, [close, value, timezone, setCurrentDate]);
 
+  const onClear = useCallback(() => setCurrentDate(undefined), [setCurrentDate]);
+
   const onApply = useCallback(
     e => {
-      if (!currentDate) {
+      if (!currentDate && disableReset) {
```
3nvi, line 119: Although I understand that you can't reset if `disableReset` is enabled (thus the `currentDate` will never turn to `undefined`), I still feel that this may lead into some bugs.

This an optional proposal, but what I'd reccommend iss cheecking the previous value of `currentDate` to decide whether you should return early  or not.

- If prev = undefined & current = undefined --> it's the first render
- If prev = something & current = undefined --> it was reset

@vorillaz thoughts on this?

src/components/DateInput/DateInput.tsx
```diff
@@ -216,11 +230,16 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
             <Button onClick={onCancel} size="medium" variantColor="gray">
               Cancel
             </Button>
-            <Button disabled={!currentDate} onClick={onApply} size="medium">
+            <Button disabled={!currentDate && disableReset} onClick={onApply} size="medium">
```
3nvi, line 233: same thoughts as above regarding the use of `disableReset` for this kind of check

src/components/DateInput/DateInput.tsx
3nvi: an implementation of this would be:

```tsx
function usePrevious<T>(value: T) {
  const ref = useRef<T>();
  
  React.useEffect(() => {
    ref.current = value;
  }, [value]);

  return ref.current;
}
```

and then do:

```tsx
const previousDate = usePrevious(currentDate)
```

3nvi (approved):
@vorillaz Thanks for this 🎉  It turned out to be more work than expected.

This is really good and it works so I'm approving 🎉 , but I'd like to discuss a bit on some of the conditionals before we proceed with a merge

vorillaz:
Lovely idea @3nvi, take a look at the implementation along with the `usePrevious` hook as it looks neat 👓 .
All the business logic is reflected and tests are also updated.

src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -115,6 +125,7 @@ const DateRangeInput: React.FC<
   const [currentMonth, setCurrentMonth] = useState(currentDateRange[0] || now(timezone));
 
   const { isOpen, open, close } = useDisclosure();
+  const previousDateRange = usePrevious(datesToDayjs(value, timezone));
```
3nvi, line 128: Wouldn't that be more correct?

```suggestion
  const previousDateRange = usePrevious(currentDateRange);
```

src/components/DateInput/DateInput.tsx
```diff
@@ -77,7 +85,19 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
   const targetRef = React.useRef(null);
   const [currentDate, setCurrentDate] = useState(dateToDayjs(value, timezone));
   const [currentMonth, setCurrentMonth] = useState(currentDate || now(timezone));
+  const resetLabel = React.useMemo(() => (withTime ? 'Clear Date & Time' : 'Clear Date'), [
+    withTime,
+  ]);
+
   const { isOpen, open, close } = useDisclosure();
+  const previousDate = usePrevious(dateToDayjs(value, timezone));
```
3nvi, line 93: Wouldn't that be more correct?

```suggestion
  const previousDate = usePrevious(currentDate);
```

src/components/DateInput/DateInput.tsx
```diff
@@ -77,7 +85,19 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
...
+
+  const isDisabled = React.useMemo(() => {
+    if (dayjs.isDayjs(currentDate) && dayjs.isDayjs(previousDate)) {
```
3nvi, line 96: wouldn't this check suffice?  This is because in my mind, `currentDate` is either a dayjs object or it's `undefined`

```suggestion
    if (currentDate) {
```

src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -197,14 +209,25 @@ const DateRangeInput: React.FC<
...
+    if (currentDateRange[0] && !currentDateRange[1]) {
+      return true;
+    }
+
+    return currentDateRange.every((date, index) => {
+      return (
+        previousDateRange &&
+        previousDateRange[index] &&
+        // @ts-ignore
+        date?.isSame(previousDateRange[index], withTime ? 'minute' : 'day')
```
3nvi, line 227: optional: You can remove the `// @ts-ignore` with this:

```suggestion
        (date as Dayjs).isSame(previousDateRange[index], withTime ? 'minute' : 'day')
```

src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -197,14 +209,25 @@ const DateRangeInput: React.FC<
...
+      return true;
+    }
+    if (currentDateRange[0] && !currentDateRange[1]) {
+      return true;
+    }
+
+    return currentDateRange.every((date, index) => {
+      return (
+        previousDateRange &&
+        previousDateRange[index] &&
```
3nvi, line 225: I  think you can fully  omit the first 2 checks. The value of `previousDateRange` is the output of  `datesToDayjs` which is always an array. Doing `date?.isSame(undefined)` will yield `false` so we are ok, no?

src/components/DateInput/ClearButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+const TextButton: React.FC<AbstractButtonProps> = props => {
+  return (
+    <AbstractButton
+      textDecoration="underline"
+      fontSize="small"
+      color="gray-300"
+      sx={{
+        '&:hover': {
+          textDecoration: 'none',
+        },
```
3nvi, line 13: optional: I don't think we have a hover behavior in the designs, but I'm not sure as well

3nvi (approved):
Thanks for that @vorillaz . I left  some comments, but I validated that the functionality  is there so I'm pre-approving 🎉 💪 

Let's make sure to address the comments before we merge :)

src/components/DateRangeInput/DateRangeInput.tsx
vorillaz: Nope, `currentDateRange` holds the input value, we need to check the validity against the controlled values and let the users decide the final outcome on "Cancel" or "Apply"

src/components/DateInput/DateInput.tsx
vorillaz: `previousDate` can be undefined and `dayjs` is failing big.

src/components/DateInput/DateInput.tsx
vorillaz: Same as above, we need to check the controlled value not the currentDate as it is changing.

src/components/DateRangeInput/DateRangeInput.tsx
vorillaz: Thanks for this one, fixed!

src/components/DateInput/ClearButton.tsx
vorillaz: That's the "TextButton" from Panther
https://github.com/panther-labs/panther-enterprise/blob/master/web/src/components/buttons/TextButton/TextButton.tsx

src/components/DateRangeInput/DateRangeInput.tsx
vorillaz: Not always, essentially we have some business logic baked in.
As users are browsing and selecting a date range they might come into a situation where they might pick just the first date from the date range, in this case, an error will be thrown. I the same manner  `previousDateRange` might be undefined throwing an error as well.

PR #213, opened 2021-04-19
Add `z-index` for modal

vorillaz:
### Background

Due to some overlapping in the panther application we need to increase the `z-index` of the modal backdrop

### Changes

- Add z-index for `[data-reach-dialog-overlay`


3nvi (approved):
Looks good! Please make sure that the following conditions are met before merging:


1. Snackbar appears above the modal
2. A tooltip on a text within the modal appears above the modal
3. A dropdown within the modal appears above the modal
4. A combobox within the modal shows its optoins above the modal

kouknick (approved):

vorillaz:
@3nvi Just checked that it works as expected.
1. Snackbar appears as expected
2., 3., 4. The `z-index` adds the stacked experience to the container thus the contents are shown as expected.

PR #214, opened 2021-04-20
Fix Combobox UI Issues

kouknick:
### Background

This PR aims to fix overflow issues in Combobox and Multicombobox components.

### Changes

- Wrap words that don't fit in Combobox
- Add padding to MultiCombobox options so the x button is always visible.

### Testing

- Locally


3nvi (approved):
Looks good Nick

PR #216, opened 2021-04-20
fix: TabList & TabPanel can accept null children

alexmylonas:
### Background

There are occasions where some tabs must be rendered conditionally thus should render null instead of children

### Changes

- Added check if child is defined before cloning

### Testing

- Used branch for local development

### DIFF
--- a/src/components/Tabs/TabList.tsx
+++ b/src/components/Tabs/TabList.tsx
@@ -14,14 +14,12 @@ export const TabList = React.forwardRef<HTMLDivElement, TabListProps>(function T
   return (
     // @ts-ignore Typing is wrong on @reach-ui, allowing only div | undefined for `as`
     <ReachTabList as={Flex} ref={ref} flexWrap="wrap" {...rest}>
-      {React.Children.toArray(children)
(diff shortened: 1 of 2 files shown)


src/components/Tabs/TabList.tsx
```diff
@@ -14,12 +14,15 @@ export const TabList = React.forwardRef<HTMLDivElement, TabListProps>(function T
...
-      )}
+      {React.Children.map(children, (child, index) => {
+        if (child) {
+          return React.cloneElement(child as React.ReactElement, {
+            isSelected: selectedIndex === index,
+            isFocused: focusedIndex === index,
+          });
+        }
+        return null;
+      })}
```
3nvi, line 25: I think `React.Children.toArray` filters out `null` values, so we could do `React.Children.toArray(children).map(....)`.

Can we try this?

alexmylonas: Yeap that seemed to also do the trick, have a look!

3nvi (approved):
Looks good @alexmylonas

PR #218, opened 2021-04-29
Collapse combobox when clicking caret

kouknick:
### Background

This Pr fixes an issue where caret arrows were not retracting the dropdown modal in Combobox components. 

### Changes

- Add onClick events in combobox arrows
###
Screencast

https://user-images.githubusercontent.com/15084305/116584569-29302200-a920-11eb-8dd9-360d3712ef66.mov

### Testing

- locally
- npm run test

### DIFF
--- a/src/components/Combobox/Combobox.tsx
+++ b/src/components/Combobox/Combobox.tsx
@@ -3,7 +3,8 @@ import React from 'react';
 import Downshift from 'downshift';
 import { filter as fuzzySearch } from 'fuzzaldrin';
 import Box from '../Box';
-import Icon from '../Icon';
+import Flex from '../Flex';
+import IconButton from '../IconButton';
 import { InputControl, InputElement, InputLabel, InputElementProps } from '../utils/Input';
 import { typedMemo } from '../../utils/helpers';
 import Menu from '../utils/Menu';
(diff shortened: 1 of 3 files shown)


src/components/Combobox/Combobox.tsx
```diff
@@ -216,16 +217,25 @@ function Combobox<Item>({
                   {label}
                 </InputLabel>
               </InputControl>
-              <Icon
+              <Flex
```
3nvi, line 220: is there a solution that doesn't involve having a button? 

Currently the whole input is clickable. I personally wouldn't want to nest a second clickable element inside another.

kouknick: Hm...I couldn't come up with a different solution. I wanted to isolate the toggle of the results only to that specific icon so the user can click as many times as he wants in the search area without toggling the results.

3nvi: In the video i see the button changing vertical alignment wheny the dropdown is opened & closed? Can we make sure that it stays in the middle of the combobox?

src/components/Combobox/Combobox.tsx
```diff
@@ -216,16 +217,25 @@ function Combobox<Item>({
...
+                justify="center"
+                pointerEvents={isOpen ? 'auto' : 'none'}
+              >
+                <IconButton
+                  tabIndex={-1}
+                  icon={isOpen ? 'caret-up' : 'caret-down'}
+                  size="medium"
+                  aria-label="Toggle Menu"
+                  onClick={() => closeMenu()}
+                  variant="unstyled"
```
3nvi, line 236: how come the unstyled variant has a background? Is that a bug?

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -368,17 +368,26 @@ function MultiCombobox<Item>({
                   {label}
                 </InputLabel>
               </InputControl>
-
               {items.length > 0 && (
-                <Icon
-                  type={isOpen ? 'caret-up' : 'caret-down'}
+                <Flex
+                  opacity={disabled ? 0.3 : 1}
```
3nvi, line 373: thanks for that!

src/components/Combobox/Combobox.tsx
kouknick: The arrow is staying at the middle of the combobox but the actual combobox height changes when the results are expanded. This happens because the clear all button is hidden when the Combobox is collapsed. This PR didn't affect the arrow position, I checked the master branch and the arrow has exactly the same behavior.

src/components/Combobox/Combobox.tsx
kouknick: I was asking myself the same thing. I tried a lot of different prop combinations in order to find a way to have a transparent background but i couldn't find one that works. I thought that you did that on purpose and it has to do with accessibility/usability compliance stuff...

3nvi: Can't remember. Could be 😆

3nvi (approved):

PR #226, opened 2021-05-07
Input number hide arrows for Firefox

vorillaz:
### Background

Firefox is a bit strict in order to hide the spinners, thus this PR uses the appearance property as it can be seen [here](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) in order to prevent showing the arrows in Firefox.

### Changes

- Change appearance global and inline rules.

### DIFF
--- a/src/components/utils/Input/InputElement.tsx
+++ b/src/components/utils/Input/InputElement.tsx
@@ -20,6 +20,7 @@ const InputElement = React.forwardRef<HTMLInputElement | HTMLTextAreaElement, In
     const theme = useTheme();
     const pt = standalone ? 14 : 5;
     const pb = standalone ? 14 : 2;
+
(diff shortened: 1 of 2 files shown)


src/components/utils/Input/InputElement.tsx
```diff
@@ -54,6 +55,7 @@ const InputElement = React.forwardRef<HTMLInputElement | HTMLTextAreaElement, In
           WebkitTextFillColor: theme.colors['gray-50'],
           borderRadius: 'medium',
         }}
+        // appearance="textfield"
```
3nvi, line 58: Shall we remove this entirely?

```suggestion

```

3nvi (approved):
Thanks @vorillaz

PR #227, opened 2021-05-10
Checkbox indeterminate state

kouknick:
### Background

Right now, the `Checkbox` component only accepts `selected: boolean`, but there are occasions i.e. when this checkbox is used as a select-all checkbox that we need to have a third "semi" selected state.

### Changes

- Add "semi" state in Checkbox component

### Screenshots
image

### Testing

- Manually

### DIFF
--- a/src/components/Checkbox/Checkbox.mdx
+++ b/src/components/Checkbox/Checkbox.mdx
@@ -4,6 +4,7 @@ const Example = () => {
   const [isCheckedB, setCheckedB] = React.useState(false);
   const [isCheckedC, setCheckedC] = React.useState(false);
   const [isCheckedD, setCheckedD] = React.useState(false);
+  const [isCheckedE, setCheckedE] = React.useState(false);
   return (
     <Box>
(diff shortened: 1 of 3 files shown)


alexmylonas (approved):
LGTM! One question that I think is not an issue, what will happen if both `checked` and `indeterminate` are true?

kouknick:
> > The checkbox will be represented in the indeterminate state.
> 
> Hmm, will be too hard to change that so that `checked=true` supersedes indeterminate? Just to protect devs when forgetting setting `indeterminate=false` when changing `checked` to truthly

I just copied how the native Html Checkboxes work. If you pass both `indeterminate=true` and `checked=true` to a native checkbox the indeterminate state is displayed. It is fairly easy to reverse this though if it fits our needs. And the more i think about it i believe you are right ;)

alexmylonas:
Practically is down to the developer to change those props correctly just feels more accurate based on our usages. Ship it! 🚢

PR #229, opened 2021-05-13
Increase tooltip stack order

kouknick:
### Background

Tooltips were rendered behind Modals. This Pr increases tooltip z-index stack order to resolve this issue.

### Changes

- Increase tooltip stack order

### Testing

- Manually


alexmylonas (approved):

3nvi (approved):
If it's not already, I'd say make it equal to modal's z-index + 1

PR #231, opened 2021-05-24
fix: correct TS types

alexmylonas:
### Background

A couple of types were misconfigured due to bumping to on reach lib.

### Changes

- Changes types for `Tab` and `PopoverTrigger`

### Testing

- Tested compilation of branch in main repo

### DIFF
--- a/src/components/Popover/PopoverTrigger.tsx
+++ b/src/components/Popover/PopoverTrigger.tsx
@@ -28,6 +28,9 @@ const PopoverTrigger = React.forwardRef<
       {...rest}
     />
   );
-}) as Polymorphic.ForwardRefComponent<'button', PopoverTriggerProps & { as: React.ElementType }>;
+}) as Polymorphic.ForwardRefComponent<
+  React.ElementType,
+  PopoverTriggerProps & { as: React.ElementType }
+>;
 
 export default React.memo(PopoverTrigger);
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tab.tsx
```diff
@@ -32,6 +33,6 @@ const Tab = (React.forwardRef<HTMLButtonElement, PrivateTabProps>(function Tab(
       {content}
     </ReachTab>
   );
-}) as unknown) as React.ForwardRefRenderFunction<'button', TabProps>;
```
3nvi, line 35: could we remove the `as unknown` or does TS complain?

src/components/Popover/PopoverTrigger.tsx
```diff
@@ -28,6 +28,9 @@ const PopoverTrigger = React.forwardRef<
       {...rest}
     />
   );
-}) as Polymorphic.ForwardRefComponent<'button', PopoverTriggerProps & { as: React.ElementType }>;
+}) as Polymorphic.ForwardRefComponent<
+  React.ElementType,
```
3nvi, line 32: shouldn't be `ElementType`, needs to be `button` here . Why are we making this change?

vorillaz (approved):

src/components/Tabs/Tab.tsx
3nvi: Gotcha. OK!

src/components/Popover/PopoverTrigger.tsx
3nvi: Ok  then :/

PR #237, opened 2021-06-02
fix: bump path-parse

alexmylonas:
### Background

Bump `path-parse` package via preinstall script

### Changes

- Added resolutions -> `path-parse: ^v1.07`

### Testing

- Install path-parse on my main branch


3nvi (approved):
LGTM

PR #244, opened 2021-06-19
Fix: make datepicker update when value updates externally

3nvi:
### Background

This PR makes sure that when `value` or `timezone` update externally, the `DateRangePicker` component updates internally as well. This helps us with Form re-initializations

### Changes

- Allow `DateRangePicker` to be updated externally
- Allow `DateInput` to be updated externally
- Fix a bug where `end date` could be before `start date`

### Testing

- `npm run test`


src/components/DateInput/DateInput.tsx
```diff
@@ -92,6 +92,12 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
   const { isOpen, open, close } = useDisclosure();
   const previousDate = usePrevious(dateToDayjs(value, timezone));
 
+  // Handles value & timezone updates outside of the component (i.e. a form has re-initialized or
+  // has updated its values as a result of an API call)
+  React.useEffect(() => {
+    setCurrentDate(dateToDayjs(value, timezone));
```
3nvi, line 98: this and the effect below will fire unnecessarily sometimes, but I think I'm unfortunately ok with that. Comparing dayjs/date/undefined to make sure it only fires when things are different seems more bug-prone to me

kouknick (approved):
Great work @3nvi! I am approving although I didn't test the changes since there are no examples in the `.mdx` of a datepicker with values being updated externally. Changes look solid though and i trust that you already tested this yourself!

3nvi:
> Great work @3nvi! I am approving although I didn't test the changes since there are no examples in the `.mdx` of a datepicker with values being updated externally. Changes look solid though and i trust that you already tested this yourself!

Yup I did. Just didn't commit the changes

vorillaz (approved):

PR #247, opened 2021-06-23
Fix: pasting JSON into the multicombo box over-tokenizes

ErikGoldman:
### Background

@willow9886 reported: https://www.loom.com/share/fce9c1ed0c7f4bb69ed7f247fb9fca5d

Pasting a JSON dictionary or array litters a Multicomboxbox with tokens. I was going to file but realized that fixing it was probably faster than putting this into the sprint board and discussing it with someone

https://app.asana.com/0/1200001670318539/1200495389667105

### Changes

Pasting a JSON dictionary pastes the keys (I don't know what the right behavior is, but it's better than what we had)

### Testing

Wrote unit tests, tested on the documentation page

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -307,9 +307,12 @@ function MultiCombobox<Item>({
                 e.stopPropagation();
 
                 // extend existing values with new ones
-                onChange(
-                  ([...items].map(processUserValue).filter(validateUserValue) as unknown) as Item[]
-                );
+                onChange([
+                  ...value,
+                  ...(([...items]
+                    .map(processUserValue)
+                    .filter(validateUserValue) as unknown) as Item[]),
+                ]);
               }
             };
 
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -99,6 +99,22 @@ export type MultiComboboxProps<T> = {
   canClearAllAfter?: number;
 };
 
+function flattenObject(ob: Array<any> | { [k: string]: any }): string[] {
+  if (Array.isArray(ob)) {
+    return ob.map(val => flattenObject(val)).flat();
+  }
+  if (ob.constructor !== Object) {
+    return [''];
+  }
+
+  const items: Array<string> = [];
+  Object.keys(ob).forEach(k => {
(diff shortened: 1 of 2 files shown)


src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -99,6 +99,22 @@ export type MultiComboboxProps<T> = {
   canClearAllAfter?: number;
 };
 
+function flattenObject(ob: Array<any> | { [k: string]: any }): string[] {
```
3nvi, line 102: I'm not sure whether the name conveys exactly what this function doees, since it creates dot-separated key paths.

I've also seen this:

```tsx
flattenObject({"x": { "y": [1,2,3]}})
// ["x.y", "x.y", "x.y"]
```

which I think is a bug.

Ultimately, if we do want this sort of behaviour, thihs should be panther-specific and not pounce-specific. What I'd propose is to allow `onPaste` or `processUserValue` to be a prop that can be customized (both would default to what we currently have) and then handle it on the panther-enterprise side.

Generally, pasting JSON into a tokenfield isn't expected, so I'm hesitent to mark this behaviour as "bug". What I do understand though, is that on a particular field in the alert summaries, you want to get a key path from a JSON. That's why I think that pounce should offer the pasting/processing customization capability and panther-enterprise should handle it as it sees fit

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -282,21 +298,37 @@ function MultiCombobox<Item>({
...
+                const parsed = JSON.parse(clipboardData);
+
+                setItems(flattenObject(parsed));
+                return;
+              } catch (e) {
+                // no error -- user input just isn't a JSON object
+              }
             }
+
+            setItems(clipboardData.replace(/\r?\n/g, ',').split(','));
```
3nvi, line 331: I feel that we should remove this business logic from herree and just allow the user to specify it. I'd even go as far as remove the default `.replace(/\r?\n/g, ',').split(','))` and just have panther-enterprise specify this on a wrapper.

Thoughts?

3nvi (commented):
Thanks for that @ErikGoldman. I think that the approach is correct, but should be defined in  panther's side. 

We should just provide the capability to allow that in pounce. What do you think?

georgesimos:
@3nvi, should we close this one?

3nvi:
I'd say so :/

PR #251, opened 2021-06-25
refactor: move from relative to absolute paths

3nvi:
### Background

This PR lays the groundwork for a small docs refactor. Moving files around is hard when you have relative paths, since you have to update a million items when moving stuff between folders.

Unfortunately, absolute paths were not an option a while ago, since the `*.d.ts` declaration files had broken paths (i.e. absolute didn't get replaced with relative). Typescript itself doesn't support it and [**won't** support it](https://github.com/Microsoft/TypeScript/issues/15479) so our options were limited or required adding a bundler like rollup or webpack just for a path transformation (something which to me seemed too convoluted).

Luckily, there's [a new package](https://github.com/justkey007/tsc-alias) that helps make this declaration file re-rewite easy! In this PR we use that and move to absolute paths so that we can easily move components around with minimal changes

### Changes

- Add re-write command in build steps
- Re-write all relative imports to absolute
- Add path aliases to `tsconfig.json`
- Add module resolutions for docs building in webpack (unrelated to the library, only related to docs)

### Testing

- `npm run validate` passes
- `npm run build` passes
- Docs are shown correctly
- Installed this branch on an existing project (panther) and validated that TS is ok there as well

### DIFF
--- a/jest/config.js
+++ b/jest/config.js
@@ -1,4 +1,13 @@
+const fromPairs = require('lodash/fromPairs');
 const { defaults } = require('jest-config');
+const tsconfig = require('../tsconfig.json');
+
+const moduleNameMapperFromTSPaths = fromPairs(
+  Object.entries(tsconfig.compilerOptions.paths).map(([alias, [aliasedPath]]) => [
+    `^${alias.replace(/\*/, '(.*)')}`,
+    `<rootDir>/${aliasedPath.replace(/\*/, '$1')}`,
+  ])
+);
 
 module.exports = {
   rootDir: '../',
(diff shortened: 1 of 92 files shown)


kouknick (approved):
I've had a quick look and it seems solid @3nvi. Any clue why the `DateRangeInput` tests are failing?

3nvi:
> I've had a quick look and it seems solid @3nvi. Any clue why the `DateRangeInput` tests are failing?

cause I fixed a TS warning wrongly

georgesimos:
@3nvi, should we close this one?

PR #256, opened 2021-07-07
Add suffix for text input

vorillaz:
### Background

When subdomains are in place we need to support a `suffix` for input text. 

### Screenshots
Screenshot 2021-07-07 at 2 08 32 PM


kouknick (approved):
Looks good!

alexmylonas (approved):
LGTM!

PR #257, opened 2021-07-08
Modify default alert props

kouknick:
### Background

 Modify default alert props

### Changes

-  Modify default alert props

### Testing

- manually


alexmylonas (approved):
LGTM! Make sure, we need this ;) And check if need to pass the old prop to existing components

PR #259, opened 2021-07-09
Downgrade `@types/react-dom` in order to play along with Panther

vorillaz:
### Background

Types are conflicting with Panther, as we have tons of TS errors from conflicting type resolutions.
image (1)

### Changes

- Downgrade types for `react-dom`

### Testing

- npm run test
- Link and play along with Panther


kouknick (approved):
LGTM!

PR #270, opened 2021-07-21
feat(combobox): add `withClearSelectionControl` prop to `Combobox`

3nvi:
### Background

We always needed a way to be able to clear your selection in Comboboxes. Currently, this happens with an `ESC` keypress but that's not so obvious for users. With this PR we add an explicit way of clearing a `Combobox`'s value

### Designs

https://www.figma.com/file/xJGJA3FyFcamwsjFd9rLIT/Styles-and-Components-Library?node-id=210%3A241

### Screencast

https://user-images.githubusercontent.com/10436045/126487401-e1739e81-2f8d-458f-b454-a067d5c5a322.mov

### Changes

- Add "Clear Selection" button
- Add `withClearSelectionControl` to enable/disable this behaviour

### Testing

- `npm run test`

### Notes

When selecting a value, the animation that closes the menu has a "flash" of green content (due to the "clear selection" button appearing as the animation for menu closing starts). I consider this an expected side-effect and I can't think of a non-hacky way to combat this

### DIFF
--- a/src/components/Combobox/Combobox.test.tsx
+++ b/src/components/Combobox/Combobox.test.tsx
@@ -27,8 +27,21 @@ const ControlledCombobox: React.FC<Partial<ComboboxProps<Item>>> = props => {
 
 describe('Combobox', () => {
   it('renders', () => {
-    const { container } = renderWithTheme(<ControlledCombobox />);
+    const { container, getByText, getByPlaceholderText } = renderWithTheme(<ControlledCombobox />);
     expect(container).toMatchSnapshot();
+
+    fireClickAndMouseEvents(getByPlaceholderText('Select manufacturer'));
+    expect(container).toMatchSnapshot();
+
+    fireClickAndMouseEvents(getByText('Toyota'));
+    expect(container).toMatchSnapshot();
+  });
+
+  it('doesn\'t render a "clear selection" button when no option is selected', () => {
+    const { queryByText, getByPlaceholderText } = renderWithTheme(<ControlledCombobox />);
+
+    fireClickAndMouseEvents(getByPlaceholderText('Select manufacturer'));
+    expect(queryByText('Clear Selection')).not.toBeInTheDocument();
   });
 
   it('clears selected value when the "clear selection" button is clicked', () => {
--- a/src/components/Combobox/Combobox.test.tsx
+++ b/src/components/Combobox/Combobox.test.tsx
@@ -1,0 +1,52 @@
+import { renderWithTheme, fireClickAndMouseEvents } from 'test-utils';
+import React from 'react';
+import Combobox, { ComboboxProps } from './index';
+
+type Item = { value: string; category: string };
+
+const items: Item[] = [
+  { value: 'Toyota', category: 'Normal' },
+  { value: 'Ford', category: 'Normal' },
+  { value: 'Chevrolet', category: 'Luxury' },
+];
+
+const ControlledCombobox: React.FC<Partial<ComboboxProps<Item>>> = props => {
+  const [selectedManufacturer, updateSelectedManufacturer] = React.useState<Item | null>(null);
+  return (
+    <Combobox
+      label="Choose a car manufacturer"
+      onChange={updateSelectedManufacturer}
+      value={selectedManufacturer}
+      placeholder="Select manufacturer"
+      items={items}
+      itemToString={item => item.value}
+      {...props}
+    />
+  );
+};
+
+describe('Combobox', () => {
+  it('renders', () => {
+    const { container } = renderWithTheme(<ControlledCombobox />);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('clears selected value when the "clear selection" button is clicked', () => {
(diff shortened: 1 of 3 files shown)


src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const selectionInput = getByPlaceholderText('Select manufacturer');
+    const selectionValue = 'Toyota';
+
+    expect(selectionInput).not.toHaveValue();
+    fireClickAndMouseEvents(selectionInput);
+    fireClickAndMouseEvents(getByText(selectionValue));
+    expect(selectionInput).toHaveValue(selectionValue);
+
+    const clearSelectionButton = getByText(
+      (_, { tagName, textContent }) => tagName === 'BUTTON' && textContent === 'Clear Selection'
```
kouknick, line 46: 👍

vorillaz, line 46: Why is `tagName` in all caps?

src/components/Combobox/Combobox.tsx
```diff
@@ -79,6 +81,9 @@ export type ComboboxProps<T> = {
 
   /** The maximum height (in pixels) of the MultiCombobox dropdown. Defaults to 300. */
   maxHeight?: number;
+
+  /** Whether a clear selection control should be available to the user */
+  withClearSelectionControl?: boolean;
```
vorillaz, line 86: Nit: `clearSelection` is better IMHO, as `withXXX` is mostly used for HOCs

src/components/Combobox/Combobox.tsx
```diff
@@ -243,6 +250,28 @@ function Combobox<Item>({
               isOpen={isOpen && results.length > 0}
               {...getMenuProps()}
             >
+              {withClearSelectionControl && (
```
vorillaz, line 253: I am not sure we should place this option as a list item, I would suggest moving the clear option outside of the list, thus we could avoid the `sticky` positioning and scroll through the actual results.

src/components/Combobox/__snapshots__/Combobox.test.tsx.snap
```diff
@@ -0,0 +1,254 @@
...
+            viewBox="0 0 24 24"
+          >
+            <path
+              d="M7 10L12 15L17 10H7Z"
+            />
+          </svg>
+        </button>
+      </div>
+    </div>
+    <ul
```
vorillaz, line 246: I would suggest firing some results rendered otherwise the snapshot is a bit off.

vorillaz (approved):
Great one @3nvi, just some minor comments and you're good to go.

src/components/Combobox/Combobox.test.tsx
3nvi: That's how the testing library exposes it. I ccan do `tagName.toLowerCase() === 'button'` if you think it's better. It's an implementation detail of the testing lib

src/components/Combobox/Combobox.tsx
3nvi: hmmm.. `clearSelection` looks like a function though. I wouldn't expect it to be a `boolean`. Also if I changed it to `enableClearSelection` it would be conceptually wrong, since user can **always** clear the sselection by pressing `ESC`. 

That's why I have used the `xxxControl` suffix, to be clear I'm referring to an element and not a functionality. How about `showClearSelectionControl`, `enableClearSelectionControl`, `displayClearSelectionControl`?

src/components/Combobox/__snapshots__/Combobox.test.tsx.snap
3nvi: ah right! I'll update it.

src/components/Combobox/Combobox.tsx
```diff
@@ -243,6 +250,28 @@ function Combobox<Item>({
...
+                <Box
+                  as="li"
```
kouknick, line 255: In the video, i noticed that this button gets hidden when the combobox is empty. I cannot see that in the code.
Also, when scrolling this button gets hidden behind the list items.

3nvi: Forgot to push it. Please re-review

kouknick (approved):
LGTM!

src/components/Combobox/Combobox.tsx
3nvi: Can you provide an example implementation? Right now the `Menu` is the one animating and containing all options. If I don't put this insidee the `Menu` component it won't animate or show correctly.

src/components/Combobox/Combobox.test.tsx
vorillaz: 👍

src/components/Combobox/Combobox.tsx
vorillaz: I would prefer the `clearSelection` approach but that's just me. It's definitely not a blocker.

vorillaz (approved):

PR #272, opened 2021-07-22
Additional card color variants

kouknick:
### Background

Add more Card color options

### Changes

- Added lighter and darker card variants

### Testing

- manually

### DIFF
--- a/src/components/Card/useCardVariantBackground.tsx
+++ b/src/components/Card/useCardVariantBackground.tsx
@@ -1,12 +1,10 @@
-import React from 'react';
 import { CardProps } from './Card';
-import { Theme } from '../../theme';
 import { BoxProps } from '../Box';
 
-type ThemeColor = keyof Theme['colors'];
+type useCardVariantBackgroundProps = Required<Pick<CardProps, 'variant'>>;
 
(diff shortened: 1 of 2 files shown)


src/components/Card/useCardVariantBackground.tsx
```diff
@@ -0,0 +1,28 @@
...
+    case 'light':
+    default:
+      return 'navyblue-400';
+  }
+};
+
+type useCardVariantBackgroundProps = Required<Pick<CardProps, 'variant'>>;
+
+const useCardVariantBackground = ({ variant }: useCardVariantBackgroundProps): BoxProps['bg'] => {
+  return React.useMemo(() => getBackgroundColor(variant), [variant]);
```
3nvi, line 25: Optional: no need for a `React.useMemo` here since you are only returning a string. We use `useMemo` to keep the samee reference on non-primitives (i.e. functions, objects & arrays).

3nvi (approved):
Great work @kouknick 🎉

src/components/Card/useCardVariantBackground.tsx
kouknick: Good call! Thanks for the tip!

PR #296, opened 2021-09-09
fix: updated props for MenuItem regarding text breaking

alexmylonas:
### Background

We are having an issue where on smaller screens (<1400px) the text on `MenuItems` is breaking words by letter.

This change seems to resolve that issue, an alternative solution to this would be to add `<Text wordWrap="nowrap">...</Text>` to every plain text that is used in a MenuItem

## Screenshot
Screenshot 2021-09-09 at 4 37 28 PM

### Changes

- Replace `wordBreak: "break-all"` with `wordWrap: "nowrap"`

### Testing

- Visual by installing branch on main repo

### DIFF
--- a/src/components/Combobox/__snapshots__/Combobox.test.tsx.snap
+++ b/src/components/Combobox/__snapshots__/Combobox.test.tsx.snap
@@ -465,7 +465,7 @@ exports[`Combobox renders 2`] = `
   transition: background-color 200ms cubic-bezier(0.0, 0, 0.2, 1) 0ms;
   position: relative;
   background-color: #3e516b;
-  word-break: break-all;
+  white-space: nowrap;
 }
 
 .pounce-8[aria-selected=true],
@@ -772,7 +772,7 @@ exports[`Combobox renders 3`] = `
   transition: background-color 200ms cubic-bezier(0.0, 0, 0.2, 1) 0ms;
   position: relative;
   background-color: #3e516b;
-  word-break: break-all;
+  white-space: nowrap;
 }
 
 .pounce-13[aria-selected=true],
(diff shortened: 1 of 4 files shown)


benjamminj (approved):
✨ 💪 💯

kouknick:
Hey @alexmylonas before we merge this can you test MultiComboBox items? We use the MenuItem component to render the ComboBox items too. With this change, large Combobox items will never break, hence they will overflow their parent.

3nvi:
@alexmylonas In your PR body you mentioned `wordSpace` but in the code changes I see `whiteSpace`. The `whiteSpace` solution is a no-go, cause it forces all menu entries to be in a single line. That won't be true in all cases as @kouknick pointed out.

I really think we should:

- either fix it in the place where we use it by adding `whiteSpace="nowrap"` to a wrapper somewhere in the panther repo where the problem lies
- fix it by adding the `minWidth` as a prop to the `DropdownMenu`

3nvi (commented):
Thanks @alexmylonas Left my comments above :)

alexmylonas (commented):
Thanks for the feedback and the recommendation guys!

@3nvi I opted for the `minWidth` to the dropdown solution, the only drawback is that I need this prop to 8 occurrences (practically to all occurences) in the main repo, I am thinking of making this a required prop and/or add a default values of `200px`

3nvi (approved):
LGTM!  @alexmylonas Would adding `minWidth="fit-content"` as the default value work? Do we still see the same issue?

Else I think I agree that adding a numbered-default is not too bad. I just feel tthat  200px is too much (especially for dropdowns only containing the options `Edit` and `Delete`). Perhaps `75px`?

alexmylonas:
@3nvi Added a `minWidth` of `75px`. It 'll catch a couple of cases

PR #299, opened 2021-09-24
fix: change title type in ControlledAlertProps interface

georgesimos:
### Background

Change ControlledAlertProps title property to `React.ReactNode` in order to avoid `@ts-ignore` when we passing title as a react node.

### Changes

- Change ControlledAlertProps title property to `React.ReactNode`

### Testing

- Manual


kouknick (approved):
🚀

alexmylonas (approved):

PR #300, opened 2021-09-27
fix: make key unique

georgesimos:
### Background

React key issue when we have the same url on more than one Breadcrumbs.

### Changes

- Add a bit more unique identifier.

### Testing

- Manual

### Screenshots
image

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -27,7 +27,7 @@ const Breadcrumbs: React.FC<BreadcrumbProps> = ({ items, ...rest }) => {
 
           return (
             <Flex
-              key={item.href}
+              key={`${item.href}_${item.text}_${index}`}
               as="li"
               alignItems="center"
               fontSize="medium"
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -27,7 +27,7 @@ const Breadcrumbs: React.FC<BreadcrumbProps> = ({ items, ...rest }) => {
 
           return (
             <Flex
-              key={`${item.href}_${item.text}_${index}`}
+              key={`${item.href}_${item.text}`}
               as="li"
               alignItems="center"
               fontSize="medium"


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -27,7 +27,7 @@ const Breadcrumbs: React.FC<BreadcrumbProps> = ({ items, ...rest }) => {
 
           return (
             <Flex
-              key={item.href}
+              key={`${item.href}_${item.text}`}
```
benjamminj, line 30: I think this will still fail the duplicate key warning if the text + href are the same like they are in your screenshot. We would need `index` in order to guarantee uniqueness unless we have an `id` field or something.

But I wonder if this would be better handled upstream in the application—why are we having the same URL appear in the breadcrumbs multiple times? I'd rather fix _that_ issue than patch the key warning inside `pounce`. Perhaps we could apply something like `_.uniq` on the breadcrumbs before we render them out.

georgesimos: We have the `Integrations` text multiple times due to react key that is not unique. (it doesn't recognize what to print correctly).
So in our scenarios  this key `{`${item.href}_${item.text}`}` should be unique. Cannot think a case that it won't.
I wanted `index` too but I removed it due to this error `error  Do not use Array index in keys  react/no-array-index-key`.
Last, we could introduce to interface `BreadcrumbItem` a new property that is not required called `id`.
And then in our Panther Enterprise we could handle the  `id`.
Also we should change the code to something like ```js  key={item.id || `${item.href}_${item.text}`} ``` for backwards compatibility. Pounce is an Open Source.

benjamminj: I guess what I'm thinking is...why are there duplicate `href` breadcrumbs in `panther-enterprise`? Is the situation that the URLs are the same but the `text` on each is different?

I would imagine we want to avoid duplicate `href` values, seems strange to have 2-3 breadcrumbs that all point to the same link.

This feels like it might be a bug upstream that we need to fix more than an issue in `pounce`.

georgesimos: By the way the issue has been resolved (check localhost).
In stage you can reproduce it by doing [those steps](https://app.asana.com/0/1200908948600021/1201063981961107/f) .

Do you think is impossible for someone to use a url with different text and we should leave the key like this?

benjamminj: Played around with it in the UI, I think I'm now seeing the scenario that we're talking about. This is because we have the "root" breadcrumb points to the same page as one of the "children" 😱 

image

I still feel like `href + text` may still be prone to future errors, I'd rather introduce that `id` field here and _fall back_ to `href`.

```diff
export interface BreadcrumbItem {
+ /** Unique identifier of the breadcrumb item */   
+ id?: string;
  /** The URL that this Breadcrumbs should navigate to when clicked */
  href: string;
  /** The text that should be displayed on this particular Breadcrumb */
  text: string;
}
```

And here:

```suggestion
              key={item.id || item.href}
```

Since we're making a change in `pounce` anyways I think we can afford to add a new field here. Then when we upgrade `pounce` we can make the little change in the `Breadcrumbs` to use the `key` as the `id`

Thoughts?

georgesimos: That is one of my thoughts so I couldn’t disagree with!
Let's see what the opinion of the rest team. 
Just for mentioning, I think this PR is optional and we can skip it!

3nvi: i honesly think that this issue has to do wth the consumer and not this package.

IMO 2 breadcrumbs should never point to the same URL. If they do, then we have an error in the way we "create" those breadcrumbs. I'd personally address it on Panther and not on Pounce

PR #301, opened 2021-10-08
feat: add large spacing values to theme spacing

benjamminj:
### Background

There's a few values in our spacing system on Figma that don't currently exist in `pounce`. This adds them as available values within `theme.spacing` so they can be used in `panther-enterprise`

There's a few values in our spacing system that _don't_ exist in the Figma mocks, but I think moving those to match the design system is a larger / more delicate migration. We can tackle that in the future.

### Changes

- Add 3 new spacing values to `theme.spacing` to match the larger Figma design system values. All of the "smaller" values already exist in `theme.spacing` (although we have some values in between)

### Testing

- `npm test`
- This is a non-breaking, addidive change so no manual testing was done. I can do a round if we want to check.

### DIFF
--- a/src/theme/theme.ts
+++ b/src/theme/theme.ts
@@ -30,7 +30,7 @@ export interface Theme extends StyledSystemTheme {
 
 export const theme: Theme = {
   breakpoints: ['769px', '1366px'],
-  space: [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40],
+  space: [0, 4, 8, 12, 16, 20, 24, 28, 32, 36, 40, 48, 64, 72],
   radii: {
     none: 0,
     small: 2,


benjamminj:
Update: after some rounds of design changes, I've actually trimmed off the designs that require these larger spacing values.
But the values still exist in our design system so I think it's a good idea for us to add them to the library.

3nvi:
Hmmm I don't know how t properly handle this.

originally we had a 10-point spacing scale with a fixed step of 4px. The [latest designs](https://www.figma.com/file/xJGJA3FyFcamwsjFd9rLIT/Styles-and-Components-Library?node-id=232%3A0) change spacing by adding a dynamic step and removing a lot of the spacings we currently use in our platform.

If wee add this change we're mixing fixed & dynamic steps in the same system. If we only keep the new ones, we'll have to make a lot of updates on pounce.

Thoughts?

cc @dimitriszoitas

benjamminj:
Gonna close this and we can tackle it at a later date.

PR #302, opened 2021-10-15
fix: add `sx` prop to `Button` typings

benjamminj:
### Background

Had some `TS` errors when using `sx` on the `Button` component. This adds the TS types to fix the error.

### Testing

- Add `sx` to `Button.text.tsx` and observe error / observed it being fixed.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -6,7 +6,9 @@ import Spinner from '../Spinner';
 import Icon, { IconProps } from '../Icon';
 import { Theme } from '../../theme';
 
-export interface ButtonProps extends NativeAttributes<'button'>, Pick<BoxProps, 'as' | 'to'> {
+export interface ButtonProps
+  extends NativeAttributes<'button'>,
+    Pick<BoxProps, 'as' | 'to' | 'sx'> {
   /** The size (height) of the button */
   size?: 'small' | 'medium' | 'large';
 


3nvi:
We normally don't want `sx` heree cause we don't want you to override the styling of buttons. Can you give me an example use case?

benjamminj:
@3nvi Here's the lines in the PR that prompted this change: https://github.com/panther-labs/panther-enterprise/pull/4624/files#diff-6a7335a419a76742c9fdc4b0e40ecfe95066d01e98adc1e044f7cdf6091045f1R43-R52

This is a case where we are making something _look_ like a `Button`, but it shouldn't be a `button` under the hood. When I switched the markup there were a couple styling things that needed tweaks (IIRC it was relying on default button layout in the browser which centers things).

benjamminj:
Another option is that we can expose some type of component that _looks_ like a `Button` but isn't interactive—however, that's the same as doing `<Button as="span" />` with this component. Which is more consistent with the way we're changing markup IMO

3nvi:
> Another option is that we can expose some type of component that _looks_ like a `Button` but isn't interactive—however, that's the same as doing `<Button as="span" />` with this component. Which is more consistent with the way we're changing markup IMO



> @3nvi Here's the lines in the PR that prompted this change: https://github.com/panther-labs/panther-enterprise/pull/4624/files#diff-6a7335a419a76742c9fdc4b0e40ecfe95066d01e98adc1e044f7cdf6091045f1R43-R52
> 
> And a comment also explaining why that markup is needed: https://github.com/panther-labs/panther-enterprise/pull/4624/files#r729342056
> 
> This is a case where we are making something _look_ like a `Button`, but it shouldn't be a `button` under the hood. When I switched the markup there were a couple styling things that needed tweaks (IIRC it was relying on default button layout in the browser which centers things).
> 
> I get the sentiment of "don't allow what we don't want" in hiding the `sx` prop, but we also want to make sure that we're exposing the functionality we need to make the UIs both pretty _and_ accessible.
> 
> On the flipside, if we'd rather `@ts-expect-error` every time we use `sx` on `Button`, we can go that route so that it highlights that usage. If that's the case I'll close this PR
> 
> Another option could be to handle the layout centering inside this component, might have downstream effects though

Left my reply [here](https://github.com/panther-labs/panther-enterprise/pull/4624/files#r730168416), but overall you got the sentimeent right. I want to help, but not open the door to **any** override.

> Another option is that we can expose some type of component that looks like a Button but isn't interactive—however, that's the same as doing <Button as="span" /> with this component. Which is more consistent with the way we're changing markup IMO

Yeah if we were using CSS that would have been a CSS class 😅 I'm open to this idea as well :) One quick way would be to expose the [internal styling hooks](https://github.com/panther-labs/pounce/blob/master/src/components/Button/useButtonStyles.tsx) and use them in panther however we want

benjamminj:
@3nvi Totally good with that approach—that way we add some customizability but with some constraints.

I think I like the following ideas you mentioned the most:
- Expose flex overrides props on the `Button`. Since that would let you control _layout_ but not something like color or text styles
- Expose the internal styling hooks. This is probably more accurate to the semantic markup in the case I linked. That allows you to drop into "power user mode" but it'll definitely not feel like the "happy path" usage of `Button`.  So it would still prompt you to answer the question "why do I need to use the raw styling utils instead of something else?"

Thoughts?

benjamminj:
@3nvi I like the idea of just making the buttons `display: flex; align-items:center` by default — like you said, I don't _think_ it'll have a huge impact. We might need `inline-flex` depending on if our buttons are set to default to block / inline.

Since this is a bigger discussion than this PR, I'm ok to close this for now and wait till we come up on the problem again to workshop some solutions.

PR #304, opened 2021-10-22
feat: add/update `Heading`, `Section` and `Article` for heading-level management

benjamminj:
## Background

One common a11y problem is skipping headings. It's best for a11y (and SEO, but that doesn't affect `panther-enterprise`) to have heading levels increase in a "semantically decreasing order", or almost as an "outline" of the page

I wanted to open this PR mainly as a starting point for a discussion of how we can solve this in our app. We don't have to go this exact route, but wanted to open the conversation 😅 

**TL;DR — this PR adds the ability to automate heading levels so that components always render the correct `h2-h6` tag**

https://user-images.githubusercontent.com/20060118/138373294-7daf4b30-3a26-44a6-85db-6996f0babd11.mov

### The Problem

When working with _components_ as the primary styling block, managing heading levels starts to get tricky. Sure, you can go look at a _page_ and see whether the headings decrease semantically, but that doesn't help make building blocks that can be placed anywhere.

Take this example component

```jsx
const UserCard = ({ user }) => {
  return (
    <div>
      <h2>{user.name}</h2>
      {/* other stuff on the card */}
    </div>
  )
}
```

If we render this component as the top-level of a page (say, like a user profile), perhaps `h2` is ok. But let's say it's embedded in another section of our page

```jsx
<main>
  <h1>Dashboard</h1>

  <section>
    <h2>Users</h2>
    {users.map(user => <UserCard key={user.id} user={user} />)}
  </section>
</main>
```

In _this case_, `h2` isn't semantically correct 😱  We'd be more correct rendering `UserCard` with a `h3`.

Many of our components that require heading levels experience this exact issue—the "proper" heading level depends entirely on _where in the DOM_ the component is rendered.

### Proposed Solution

Turns out, `React.Context` can save our bacon and automate 90% of this problem away.

This PR adds two components, `H` and `Section`. Both components work _together_ to automate heading levels so that an `H` tag always displays the correct heading level regardless of where it's rendered.

```jsx
const UserCard = ({ user }) => {
  return (
    <div>
      <H>{user.name}</H>
      {/* other stuff on the card */}
    </div>
  )
}

// Dashboard
<main>
  {/** h1 doesn't need to be adaptive, there should only ever be 1 on the page */}
  <h1>Dashboard</h1>

  <section>
    {/* _Technically don't need <H> here since it's a page-level component. But this renders a `h2` */}
    <H>Users</H>
    {/* This currently adds another "section" tag */}
    <Section>
      {users.map(user => <UserCard key={user.id} user={user} />)}
    </Section>
  </section>
</main>
```

These two components allow you to drop `H` in wherever headings makes sense _semantically_ without needing to worry where the component is rendered. And then parent components (or layout components) can set up `<Section>` wrappers to make sure that the `h1-h6` progression is correct!

_Believe it or not, this exact behavior was technically in [the HTML5 spec](https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements) for `h` and `section`, but is not recommended bc no browsers / screenreaders fully implemented it!_

### Rollout plan
1. Add `H` and `Section` to `pounce`. Convert all `pounce` components to not manually set heading levels.
2. Install new version in `panther-enterprise`. Some heading levels will change in the underlying markup, but it will be very few (mostly the default `h4` tags becoming `h2` tags)
3. Audit `panther-enterprise` and switch any instances of `as="h3"` (and other levels) to use `as={H}`. This should also be a simple search-and-replace.
4. Add `Section` components to the page layouts to allow the `H` tags to set the proper contextual level. This will be a little bit more case-by-case since `Section` has underlying markup so we'll need to check that we're not breaking things. IMO this can be done at the same time as Step 3

### Resources

- https://web.dev/heading-order
- https://dequeuniversity.com/rules/axe/4.2/heading-order
- https://css-tricks.com/managing-heading-levels-in-design-systems/
- https://medium.com/@Heydon/managing-heading-levels-in-design-systems-18be9a746fa3 (primary source of this approach)

## Changes

- Add `H` and `Section` components
- Change all components using `as="h4"` (or other levels) to `as={H}` so they will get contextual heading levels.

## Testing

- `npm test`
- Manually inspecting HTML in Styleguidist examples


kouknick (commented):
Thanks for that @benjamminj 💯 ! That's a really interesting approach to deal with the Headings levels. We will be adding some context overhead by using a number of nested providers but since the providers only return a static number there won't be any issue! 
Deferring to @3nvi for approval

3nvi (commented):
Great work here. I like the approach, let's nail down the the API and see if we can do something about multiple `h1` tags (no `h1` siblings should ever exist unless I'm mistaken)

src/components/H/Section.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const level = React.useContext(HeadingLevelContext);
+
+  return (
+    <HeadingLevelContext.Provider value={level + 1}>
+      <Box {...rest} as={as}>
+        {children}
+      </Box>
+    </HeadingLevelContext.Provider>
+  );
+};
```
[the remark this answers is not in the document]
benjamminj: Yeah I dig that. Section + Article are the 99% use case where you're gonna increase heading levels. And if you wanna do it without the `section` / `article` markup, there's always `<Section as="div" />`

src/components/H/H.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Provides and manages `h` tags within a document context. The underlying heading level rendered
+ * can vary depending on _where_ this component is in the component tree.
+ *
+ * This allows for simple scaffolding of a document with accessible heading levels while
+ * also allowing components to create markup in a vacuum.
+ */
+export const H: React.FC<HProps> = React.forwardRef<HTMLHeadingElement, HProps>(function H(
+  { increase = 0, ...rest },
```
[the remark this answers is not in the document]
benjamminj: I don't think naming the prop `level` portrays the right mental model—not if we want these tags to automatically manage levels properly. The primary reason being that `level` implies the absolute level, rather than the rest of the component operating with a _relative awareness_ of its level.

```tsx
// Let's assume we're at the h4 level in the ctx.
// I'd expect this to output an `h2` by a simple reading of the component.
<H level={2}>Heading</H>
```

I wonder if we can axe `level` or `increase` entirely? The situations where we would need two adjacent heading levels are decently rare, especially since we don't have SEO as a main concern. In the off-chance that we do need adjacent headings, you can always wrap the heading in a new section to increase the level.

```tsx
<H>Heading 2</H>
<Section as='span'>
  <H>Heading 3</H>
</Section>
```

benjamminj:
@3nvi I'm not 100% sure if we _need_ to handle `h1` tags as part of _this_ automated heading level management — that was kinda my thought in only allowing `H` to start at 2.

The constraints on `h1` are a little bit stricter than `h2-h6` since you also need to make sure that there's only 1 per page (at most 2, according to [this screen reader user survey](https://webaim.org/projects/screenreadersurvey7/)). While it's _technically ok in terms of SEO / HTML, seems like it's a better practice to only have 1 per page that's the page title.

I'm almost wondering if we should just have a separate `H1` component (perhaps more abstract, like `PageTitle`) that had _very strict constraints_ and threw dev errors / warnings. Similar to how React warns about missing keys in dev but not in prod. For example:

```
const H1 = () => {
  // check right here for any other h1 tags in the document, but only in development, we don't wanna
  // blow up prod over this.
  return <h1>{...}</h1>
}
```

Another option would be to bake this into our root-level layout in `panther-enterprise`—we already do things like setting the page title there so I could just imagine an `h1` inline there and we don't need to solve `h1` tags in `pounce`. Might be the path of least resistance

benjamminj:
@3nvi @kouknick Had some time to circle back around to this, did a round of edits.

1. Removed the `increase` prop entirely. I think we can get by without for right now, what's mainly important is that `h2-h6` are automated and we don't have any locations (to my knowledge) of adjacent heading levels that would need the `increase` prop.
2. Added `Article` and `Section` as separate components. Functionally they do the same thing, they just render different markup under the hood.

And a couple notes on where I think we can go with this:
- I don't think we should bake `h1` handling into this component—the constraints and validations would need to be different than `h2-h6`. But I can explore that in a separate PR, perhaps we use a different component.
- We _could_ expose this as a hook in addition (or instead of the `H` component). But personally I'm a fan of composing the `H` component—since we wouldn't be limited by rules of hooks.

src/components/Article/Article.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { BoxProps } from '../Box';
+import Section from '../Section';
+
+/**
+ * Increases the heading level of all `H` components inside it by 1.
+ *
+ * NOTE: this component is functionally the same as <Section />, but semantically
+ * it renders an <article> tag instead of a <section> tag.
+ */
+export const Article: React.FC<BoxProps<'article'>> = ({ as = 'article', ...rest }) => {
```
benjamminj, line 11: I imagine these need their own example `mdx` files too? Will be essentially the same as the `H` example

3nvi: Yeah. every `components/{COMPONENT}` folder needs:

- a test file
- an mdx file for docs (even if minimal)

src/components/H/H.tsx
```diff
@@ -0,0 +1,28 @@
...
+export type HProps = BoxProps<HTag>;
+
+/**
+ * Provides and manages `h` tags within a document context. The underlying heading level rendered
+ * can vary depending on _where_ this component is in the component tree.
+ *
+ * This allows for simple scaffolding of a document with accessible heading levels while
+ * also allowing components to create markup in a vacuum.
+ */
+export const H: React.FC<HProps> = React.forwardRef<HTMLHeadingElement, HProps>(function H(
```
3nvi, line 17: question: why don't we replace our existing `Heading` with this? I don't see any reasson to have both:

- `H`
- `Heading`

As a dev, I'm a bit confused by the presence of both. Are there any downsides (apart from the context reading which is fine) that you see?

P.S. We'll have to handle `level` being `undefined` in that case, but that's easy to fix

src/components/H/H.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Section>
+        <H>h3</H>
+      </Section>
+    );
+
+    expect(container.querySelector('h2')).not.toBeInTheDocument();
+    expect(getByText('h3').tagName).toEqual('H3');
+  });
+
+  it('does not render above an h6', () => {
```
3nvi, line 23: can we move some of these tests to `components/Section/Section.test.tsx` and `components/Article/Article.test.tsx` accordingly?

src/components/Section/Section.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Box, { BoxProps } from '../Box';
+import { HeadingLevelContext } from '../H';
+
+/**
+ * Increases the heading level of all `H` components inside it by 1.
+ */
+export const Section: React.FC<BoxProps<'section'>> = ({ children, as = 'section', ...rest }) => {
+  const level = React.useContext(HeadingLevelContext);
```
3nvi, line 9: q: I see that we have a default context value, but I'm a bit baffled by something. Can't `level` never be `undefined` if there's no parent provider? 

I see that the related tests pass, but I'm failing to understaand why. In my mind useContext returns `undefined` if  no provider parent exists, but I see that somehow in your tests that's not the case

3nvi (commented):
I really like the overall approach @benjamminj . Things that I need to approve:

- have `Section` and `Article` have a test & mdx file
- help me understand why `level` can **never** be `undefined` (as per my comments)
- discuss whether `H` can replace `Heading` and provide evidence as to why this would be wrong

Sorry for being pedantic here, I just want to make sure we keep an "easy to understand" API :)

src/components/H/H.tsx
```diff
@@ -0,0 +1,47 @@
...
+  { increase = 0, ...rest },
+  ref
+) {
+  const level = React.useContext(HeadingLevelContext);
+
+  // `Math.min` makes sure that the heading level is never above a h6
+  const tag = 'h' + Math.min(level + increase, 6);
+  return <Box ref={ref} {...rest} as={tag as HTag} />;
+});
```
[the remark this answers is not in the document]
3nvi: > Heading comes with styles baked in, and currently its main responsibility is providing those heading styles.

I kind of disagree with that. It has some default styles like the browser's default stylesheet has, but not something that  you have to continuously override. The **only** thing it has as a default is the size. Literally nothing else.

> Feels like we'd be starting to mix the "concerns" of Heading rather than composing a few smaller pieces together

I understand. My mindset behind that is "I don't want to have to think if I should use `H` or `Heading`". Generally, I wouldn't want to have multiple ways off exposing the same thing, and I'd hate to see `<H as={Heading} />` since it just feels weird.

Should we chat in a small call?

src/components/Section/Section.tsx
benjamminj: > Can't level never be undefined if there's no parent provider?

Nope! The default value to `createContext(1)` takes care of that!

> In my mind useContext returns undefined if no provider parent exists

This is slightly off from the way `Context.Consumer` & `useContext` works. If the context is given a default value, that will be provided by `useContext` in the event that it's rendered without a provider.

[React docs:](https://reactjs.org/docs/context.html#reactcreatecontext)

> The defaultValue argument is only used when a component does not have a matching Provider above it in the tree.

benjamminj:
@3nvi I finally got around to another round of edits on this PR while waiting on a `mage teardown` 😅 

> have Section and Article have a test & mdx file

Done! Each has a test / mdx file.

> help me understand why level can never be undefined (as per my comments)

Explanation [here](https://github.com/panther-labs/pounce/pull/304#discussion_r790073242), level will never be undefined even if the headings are rendered outside of a provider.

> discuss whether H can replace Heading and provide evidence as to why this would be wrong

I was able to roll `H` _into Heading entirely_, which means that this feature will "just work" out of the box with our existing components and there's no breaking changes to the existing component interfaces. This should make migration in `panther-enterprise` a lot easier.

src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -313,10 +314,10 @@ const DateRangeInput: React.FC<
                   aria-label="Go to previous month"
                 />
               </Box>
-              <Box as="h4" fontSize="medium" fontWeight="bold" width="50%" textAlign="center">
+              <Box as={Heading} fontSize="medium" fontWeight="bold" width="50%" textAlign="center">
                 {currentMonth.format('MMMM YYYY')}
               </Box>
-              <Box as="h4" fontSize="medium" fontWeight="bold" width="50%" textAlign="center">
+              <Box as={Heading} fontSize="medium" fontWeight="bold" width="50%" textAlign="center">
```
benjamminj, line 320: No visual change here:

image

src/components/DateInput/DateInput.tsx
```diff
@@ -197,7 +198,7 @@ const DateInput: React.FC<DateInputProps & Omit<TextInputProps, 'value' | 'onCha
             icon="arrow-back"
             aria-label="Go to previous month"
           />
-          <Box as="h4" fontSize="medium" fontWeight="bold" tabIndex="-1">
+          <Box as={Heading} fontSize="medium" fontWeight="bold" tabIndex="-1">
```
benjamminj, line 201: image

src/components/Modal/Modal.tsx
```diff
@@ -69,7 +69,7 @@ const Modal: React.FC<ModalProps> = ({
                   <Card minWidth="400px" position="relative" boxShadow="dark200">
                     {title && (
                       <Box as="header" borderBottom="1px solid" borderColor="navyblue-300" py={6}>
-                        <Heading as="h4" size="x-small" textAlign="center" id={id}>
+                        <Heading as={Heading} size="x-small" textAlign="center" id={id}>
```
benjamminj, line 72: image

kouknick: Heading as heading :P

kouknick (commented):
I think that we can merge this. Deferring approval to @3nvi  since he had some suggestions/comments but i think that all of them are already addressed. Great job there!

src/components/Modal/Modal.tsx
benjamminj: 😂 

We might even be able to remove this prop entirely now...I'll give it a try

benjamminj:
@3nvi have a couple seconds to take a peek at this? Hoping to close this one out before I go on leave 😅

src/components/Heading/Heading.tsx
```diff
@@ -2,7 +2,15 @@ import React from 'react';
...
 import useHeadingStyles from './useHeadingStyles';
 
-export interface HeadingProps extends BoxProps<'h1'> {
+// Default heading levels to `h2`:
+//
+// This is so that you can drop `<Heading>` on a page _anywhere_ and it will be
+// semantically correct. Since you should _usually_ only have 1 `h1` on the page,
+// that can be done by explicitly overriding the `as` prop (`<Heading as="h1">`).
+const DEFAULT_HEADING_LEVEL = 2;
+export const HeadingLevelContext = React.createContext(DEFAULT_HEADING_LEVEL);
```
georgesimos, line 11: [question] who is using or modifying this context?

src/components/Section/Section.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Box, { BoxProps } from '../Box';
+import { HeadingLevelContext } from '../Heading';
+
+/**
+ * Increases the heading level of all `Heading` components inside it by 1.
+ */
+export const Section: React.FC<BoxProps<'section'>> = ({ children, as = 'section', ...rest }) => {
+  const level = React.useContext(HeadingLevelContext);
```
georgesimos, line 9: There it's! 
Also, I don't get why we have to increase the containing headings by 1? 
Sections should always have a heading, with very few exceptions. 
So an `h1` -> `h6` will do the job, but why don't we use an h2 over h3? 

I think it would be better to default the `hTag` to `h2` and don't use the context at all!

src/components/Heading/Heading.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  it('allows rendering a h1', () => {
+    const { getByText } = renderWithTheme(<Heading as="h1">Hello</Heading>);
+    expect(getByText('Hello').tagName).toEqual('H1');
+  });
+
+  it('does not render above an h6', () => {
+    const { getByText } = renderWithTheme(
+      <div>
+        <Heading>h2</Heading>
+        <Section>
```
georgesimos, line 21: I thought `section` represents a generic standalone section of a document and not that during nested sections we should increase the heading level!

3nvi: I think that was the whole point of this PR 😄  That you don't have to think about `h1` vs `h2`, but that Pounce would do that for you

src/components/Modal/Modal.tsx
3nvi: `<Heading as={Heading} ` seems completely wrong 😄  Can we not remove the `as`? It would render an `h2`, right?

3nvi (commented):
Thanks @benjamminj! I'm ok to approve as long as we change the default heading inside a section :)

src/components/Modal/Modal.tsx
benjamminj: Yeah this was leftover from one of the previous versions, removed it 😂

3nvi (approved):
After reading the comments and the last updates, it LGTM 🎉 

Sorry for taking that long 🙏

PR #306, opened 2021-11-15
feat: change text from string to ReactNode

georgesimos:
### Background

Ability to pass a react node instead of only string in breadcrumb item title!

### Changes

- [x] Modify text  shape
- [x] Remove unused part from mdx

### Testing

- Manual

### Screenshot

Screenshot 2021-11-15 at 4 44 38 PM

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -7,19 +7,14 @@ import Link, { LinkProps } from '../Link';
 export interface BreadcrumbItem {
   /** The URL that this Breadcrumbs should navigate to when clicked */
   href: string;
-  /** Title is the text that should be displayed on this particular Breadcrumb
-   * element can be: numbers, strings, elements or an array (or fragment) containing these types.
+  /** The react node that will be displayed on this particular Breadcrumb
+   * children can be: numbers, strings, elements or an array (or fragment) containing these types.
    */
-  title: React.ReactNode;
-  /**
-   * @deprecated use node property instead.
-   * The text that should be displayed on this particular Breadcrumb
-   */
-  text?: string;
+  children: React.ReactNode;
 }
 
 export interface BreadcrumbProps extends Pick<LinkProps, 'as'> {
-  /** A list of `BreadcrumbsItem` objects ( `{href,title,text}` ) that will construct the Breadcrumb */
+  /** A list of `BreadcrumbsItem` objects ( `{href,children}` ) that will construct the Breadcrumb */
   items: BreadcrumbItem[];
 }
 
@@ -49,8 +44,7 @@ const Breadcrumbs: React.FC<BreadcrumbProps> = ({ items, ...rest }) => {
                 data-active={isLastBreadcrumb ? true : undefined}
                 {...rest}
               >
-                {/* TODO: remove item.text  */}
-                {item.title || item.text}
+                {item.children}
               </Link>
               {!isLastBreadcrumb && (
                 <Icon
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -7,9 +7,10 @@ import Link, { LinkProps } from '../Link';
 export interface BreadcrumbItem {
   /** The URL that this Breadcrumbs should navigate to when clicked */
   href: string;
-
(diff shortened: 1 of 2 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -7,9 +7,10 @@ import Link, { LinkProps } from '../Link';
 export interface BreadcrumbItem {
   /** The URL that this Breadcrumbs should navigate to when clicked */
   href: string;
-
-  /** The text that should be displayed on this particular Breadcrumb */
-  text: string;
+  /** The text that should be displayed on this particular Breadcrumb
+   * text can be: numbers, strings, elements or an array (or fragment) containing these types.
+   */
+  text: React.ReactNode;
```
georgesimos, line 13: I could also point text as deprecated and create a new property called node or element.
Thoughts?

3nvi: Yeah I like this idea. Let's go for that :)

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -43,7 +49,8 @@ const Breadcrumbs: React.FC<BreadcrumbProps> = ({ items, ...rest }) => {
                 data-active={isLastBreadcrumb ? true : undefined}
                 {...rest}
               >
-                {item.text}
+                {/* TODO: remove item.text  */}
```
benjamminj, line 52: Do we have a target version of `pounce` where we intend to remove `text`? Perhaps we should make an Asana task to migrate `panther-enterprise` and fully remove the value

georgesimos: I agree, usually if someone is using this product you shouldn't change properties like this. 
You are doing this in phases!

But you are welcome to provide any thought on this!

benjamminj: Yeah totally. I think as long as we add an Asana task to migrate `panther-enterprise` and remove this prop I'm good.

Since we only have 1 app using `pounce` and we're still a `0.x.x` version I'm ok with the breaking change, as long as we provide a migration path.

kouknick (approved):
🥂

benjamminj (approved):
🔥

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -7,13 +7,19 @@ import Link, { LinkProps } from '../Link';
...
   href: string;
-
-  /** The text that should be displayed on this particular Breadcrumb */
-  text: string;
+  /** Title is the text that should be displayed on this particular Breadcrumb
+   * element can be: numbers, strings, elements or an array (or fragment) containing these types.
+   */
+  title: React.ReactNode;
+  /**
+   * @deprecated use node property instead.
```
3nvi, line 15: This commentt is  wrong, since `node` doesn't exist

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -7,13 +7,19 @@ import Link, { LinkProps } from '../Link';
 export interface BreadcrumbItem {
   /** The URL that this Breadcrumbs should navigate to when clicked */
   href: string;
-
-  /** The text that should be displayed on this particular Breadcrumb */
-  text: string;
+  /** Title is the text that should be displayed on this particular Breadcrumb
+   * element can be: numbers, strings, elements or an array (or fragment) containing these types.
+   */
+  title: React.ReactNode;
```
3nvi, line 13: Since a breadcrumb may have links, dropdowns, etc., I feel that `node` or `children` is better.  I actually think I prefer `children` unless you don't

src/components/Breadcrumbs/Breadcrumbs.tsx
georgesimos: sorry renamed to title!

src/components/Breadcrumbs/Breadcrumbs.tsx
georgesimos: No I like it, but don't you think this is going to overwrite the children prop from react? 
Do you like the `render` term?
It will be like render props without logic!

3nvi (approved):
LGTM 🎉

PR #307, opened 2021-11-18
Number input callback dependencies

kouknick:
### Background

This PR fixes a bug in our NumberInput component where the `stepUp` and `stepDown` callbacks were not updating when their dependencies get modified. 

### Changes

- fix the dependency issue

### Testing

- manually


3nvi (approved):
Thanks for the fix! LGTM

PR #310, opened 2021-12-02
feat: change color theme based on figma

georgesimos:
### Background

Modify colors palette based on figma desings : https://www.figma.com/file/xJGJA3FyFcamwsjFd9rLIT/Styles-and-Components-Library?node-id=0%3A1

### Changes

- [x] Convert color's values to uppercase
- [x] Replace gray-50 to white-200 @dimitriszoitas Am I right? 
- [x] Change default button variantColor to blue-300

### Testing

- Modify snapshots

#### Before & After Button

Screenshot 2021-12-02 at 1 11 28 PM
Screenshot 2021-12-02 at 1 11 50 PM


src/theme/colors.ts
```diff
@@ -2,138 +2,139 @@ const colors = {
...
-  'navyblue-200': '#4f627c',
-  'navyblue-300': '#3e516b',
-  'navyblue-350': '#30415C',
-  'navyblue-400': '#23344e',
-  'navyblue-500': '#1d2a3f',
-  'navyblue-600': '#192231',
-  'navyblue-700': '#141c2a',
-  'navyblue-800': '#0f1623',
-
-  'gray-50': '#f6f6f6',
```
benjamminj, line 20: I'm ok with the capitalization / alphabetical, but I don't think this PR should do changes on the color system itself. Can we undo the breaking changes to color tokens and keep this PR scoped only to button variants?

The button change is a small, low-risk change, but changing the color tokens will require migration steps in `panther-enterprise` and likely more discussion.

benjamminj (dismissed):
Can we scope this PR down to _only the color change on `Button`_? That is a non-breaking change, easy to review, and requires 0 migration steps when we bump the version in `panther-enterprise`.

We should keep the changes to the color tokens separate, that is a breaking change and would require migrating the color tokens in `panther-enterprise`. I also anticipate that PR will require more discussion.

src/theme/colors.ts
kouknick: I don't know if this will require a lot of migration steps. It will be a simple mass replacement from gray-50 to white-300 but i cannot see the white-300 in the Figma color pallet either so i don't know if this change is required or not.

georgesimos: I have asked the design team ( Jim ) about white-*. And I will come back to you!

@benjamminj this task is about the colors and not the button ( Added because this PR looking a good place for this change - We could open another one too). 

The change on the format (Capital) was because it will be easier to copy-paste the color value from Figma to pounce. 

I think this PR doesn't have any breaking changes and the migration steps doesn’t require any big effort.
In any case if you aware of this we could close the PR. ( I made this PR based [on this thread](https://panther-labs.slack.com/archives/C02NVPZ1ZL4/p1638386009011000)! )

 [Created task ](https://app.asana.com/0/1200908948600021/1201451000624090/f) for the migration.

benjamminj: If you're good to do the migration of tokens after we publish, I'm ok with that.

Agree with @kouknick, not sure about `white-300`, I only see `white-100` and `white-200` in the figma palette

georgesimos: So after a discussion I had with Jim, the gray-50 is going to be replaced by white-200!
The migration step is just a simply replacement!

src/components/Tabs/__snapshots__/Tabs.test.tsx.snap
```diff
@@ -15,7 +15,7 @@ exports[`Tabs renders 1`] = `
 .pounce-0 {
   z-index: 1;
   cursor: pointer;
-  color: gray-50;
+  color: white-200;
```
benjamminj, line 18: It's strange that this snapshot is rendering the actual name of the color rather than the hex code. 

I don't think we should block this PR for it, it's been doing that since before this PR

benjamminj (approved):
Looks great!! Thanks @georgesimos 🎉

kouknick (approved):
LGTM

PR #317, opened 2022-01-22
fix: switch focus state and screenreader output

benjamminj:
### Background

This fixes two bugs found while working with `<Switch />`

#### 1. Switch screenreader is too verbose

[Asana ticket](https://app.asana.com/0/1200879509676555/1201620510067156/f)

The switch component currently reads the label, text, _and_ checkbox state. However, since the screenreader already reads the checkbox state the "On" / "Off" checkbox texts are more decorative.

It's a slight improvement—wasn't broken before, but this is a nicer UX when you're using a screenreader. That way the label you're focused on doesn't change while you interact with it (just the checked/unchecked state)

- Before: `With Custom names NO unchecked checkbox`
- After: `With Custom names unchecked checkbox`

#### 2. No focus state

The checkbox didn't have a focus state!! I added one in, just a white border for now so that it's accessible.

Before, focus on the checkbox:

Screen Shot 2022-01-21 at 5 18 45 PM

After, focus on the checkbox:
Screen Shot 2022-01-21 at 6 01 54 PM

### Testing

- Added a unit test for the label.

### Implications in `panther-enterprise`

For the most part, the actual components should be good to go, no breaking changes. However, I'll need to update any _tests_ that were relying on the label text. I'm happy to own this on panther-enterprise.


src/components/Switch/Switch.tsx
```diff
@@ -19,6 +20,12 @@ export type SwitchProps = NativeAttributes<'input'> & {
   uncheckedText?: string;
 };
 
+const switchFocusStyles = {
+  '& .pill': {
```
benjamminj, line 24: There's probably a more CSS-in-JS way to select the nested pill than a CSS class, not sure. I've always found this to be tricky and confusing in Emotion / Styled Components / etc, even though the CSS itself is simple

georgesimos (commented):
LGTM, do we have the ok from the Design team?

src/components/Switch/Switch.tsx
3nvi: Why don't we simply put the `_focusWithin` on the `Flex` component itself? Tested it and it works just fine + we avoid the need for a CSS selector?

src/components/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from 'test-utils';
+import Switch from './Switch';
+
+describe('Switch', () => {
```
3nvi, line 5: minor: while running the test I saw a warning around `getComputedStyle`. Can we include this in `jest/config.js` in order to make the warnings go away?

```js
// jest-dom does not properly implement this method and this workaround
// keeps the test runner from throwing warnings
const { getComputedStyle } = window;
window.getComputedStyle = elt => getComputedStyle(elt);
```

3nvi (commented):
Great initiative on this! I'm fine to approve as is, but I was wondering if we can move the focus rule one component deeper.

While there, do you have idea how could we implement the `Switch` component in a simple way so that `getByLabelText('the switch label')` would work? Currently it returns nothing due to the nested `<span>` where the text lives 😢

benjamminj:
> While there, do you have idea how could we implement the Switch component in a simple way so that getByLabelText('the switch label') would work? Currently it returns nothing due to the nested <span> where the text lives 😢

I think I have some ideas, I'll play around with them when I get a little time. We might be able to slightly tweak the HTML / CSS which will clean up the label in testing

benjamminj:
> LGTM, do we have the ok from the Design team?

@georgesimos nice call, I brought it up in Slack, let's see what the design folks think

benjamminj:
Alrighty, just pushed a wave of updates that addresses all the feedback (thanks @3nvi @georgesimos!!)

- synced with design and updated the focus state to `blue-200` per their feedback
  Screen Shot 2022-01-26 at 11 19 45 AM
  Screen Shot 2022-01-26 at 11 19 36 AM
- reworked HTML to make test selection easier. This also meant we don't need to do `aria-hidden` anymore since the label is now the inner label element and attached to the input via `id / htmlFor`
- fixed the test mock in `setup.ts` for `getComputedStyle`
- move focus styles deeper on the UI itself rather than container to avoid nested CSS rule.

georgesimos (approved):
LGTM!

benjamminj:
Gonna merge this, I'll own upgrading pounce & propagating changes to `panther-enterprise`

PR #318, opened 2022-01-28
Fix date range validation

kouknick:
### Background

Fixes validation on DateRange picker to ensure that the end date is always after the start date.

### Changes

- Added start date < end date validation in date range picker 

### Testing

- npm run test

### DIFF
--- a/src/components/DateRangeInput/DateRangeInput.tsx
+++ b/src/components/DateRangeInput/DateRangeInput.tsx
@@ -230,7 +230,11 @@ const DateRangeInput: React.FC<
     if (areDatesUndefined(previousDateRange) && areDatesUndefined(currentDateRange)) {
       return true;
     }
-    if (currentDateRange[0] && !currentDateRange[1]) {
+    if (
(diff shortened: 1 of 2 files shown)


src/components/DateRangeInput/DateRangeInput.tsx
```diff
@@ -230,7 +230,11 @@ const DateRangeInput: React.FC<
     if (areDatesUndefined(previousDateRange) && areDatesUndefined(currentDateRange)) {
       return true;
     }
-    if (currentDateRange[0] && !currentDateRange[1]) {
+    if (
+      currentDateRange[0] &&
```
kouknick, line 234: This might seem a bit off. It took me a while to realize why we check if the currentDateRange[0] must have a value. This is simply because we have the "Clear dates" button where a user can click to reset the dates and then he can click on the save button to reset the component values.

3nvi (approved):
Looks good @kouknick 

Hopefully we can update all of this code at one point and make it a bit more maintainable 😢

PR #319, opened 2022-02-08
fix: make click target on Switch larger

benjamminj:
### Background

Some of the semantic changes to fix the `Switch` component's accessible labels decreased its clickable area. This patch increases the input's clickable area to make sure that `Switch` is clickable in the entire pill

### Before

https://user-images.githubusercontent.com/20060118/152894567-83a0259d-be50-4523-9d0e-bfe57bc14acb.mov

### After

https://user-images.githubusercontent.com/20060118/152894617-7af08a57-d90c-471f-b9ed-2c8cb76dc4dc.mov


3nvi (approved):
Amazing @benjamminj ! That was my feedback around clickable labels on switched, that if you add proper `<label>` semantics you are missing the clicking area.

Wanted to comment that on your last PR but I forgot to & it got merged. It's nice to know that we found a way around it. Props 🎉

georgesimos (approved):

src/components/Switch/Switch.tsx
```diff
@@ -105,6 +105,10 @@ const Switch = React.forwardRef<HTMLInputElement, SwitchProps>(function Switch(
           as="input"
           cursor="pointer"
           position="absolute"
+          height="100%"
+          width="100%"
+          top={0}
+          left={0}
```
3nvi, line 111: @benjamminj after my testing I saw that we also need this

```suggestion
          left={0}
          margin="auto"
```


Without margin:
Screenshot 2022-02-09 at 6 39 42 PM

With margin:
Screenshot 2022-02-09 at 6 39 46 PM

benjamminj: cool, I'll add that in

PR #330, opened 2022-03-08
feat: add show-password icon

benjamminj:
### Background

Adds a new icon from Figma, it's named "show password" there


3nvi (approved):
Approving, although we try and use realistic naming. For example, if something is an `eye` icon, we'd call it `eye` even if it's used for passwords.

benjamminj:
> Approving, although we try and use realistic naming. For example, if something is an eye icon, we'd call it eye even if it's used for passwords.

IMO it's valuable for us to call things the same thing in Figma and `pouncejs` — reduces the friction for developers while using `pouncejs`. I'll ask design about renaming these and see what they think

3nvi:
I agree. If too much work is involved, let's proceed as is

PR #331, opened 2022-04-04
feat: expand multicombobox props

shaun-sweet:
### Background

Looking to make an alteration to the `MultiComboBox` that matches this figma example - https://www.figma.com/proto/m8HKEVhIUX0qGaIB0geBVx/Data-Explorer-V2?page-id=0%3A1&node-id=87%3A6959&viewport=352%2C48%2C0.31&scaling=min-zoom&starting-point-node-id=84%3A5970

### Changes

- Added `renderPlaceHolder` prop that allows you to pass in a custom placeholder
- Changed default `Clear All` banner button to the new design with the pleasant green icon 
image

### Testing

- Added one test to verify custom placeholder passes properly
- Updated snapshots

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.mdx
+++ b/src/components/MultiCombobox/MultiCombobox.mdx
@@ -180,7 +180,7 @@ const Example5 = () => {
 <Example5 />;
 ```
 
-A MultiCombobox can have a custom placeholder passed to it
+A MultiCombobox can have custom content passed to it
 
 ```typescript jsx
 const Example6 = () => {
@@ -196,7 +196,7 @@ const Example6 = () => {
         <MultiCombobox
           label="Filter columns"
           variant="solid"
-          renderPlaceHolder={() => <div>Filtered Columns ({selectedMoto.length})</div>}
+          renderContent={() => <div>Filtered Columns ({selectedMoto.length})</div>}
           items={['Benelli', 'Ducati', 'BMW', 'Moto Guzzi', 'Piaggio']}
           hideLabel
           onChange={updateSelectedMoto}
--- a/src/components/MultiCombobox/MultiCombobox.mdx
+++ b/src/components/MultiCombobox/MultiCombobox.mdx
@@ -216,17 +216,22 @@ A MultiCombobox can render a custom clear all selected element
 
 ```typescript jsx
 const Example7 = () => {
-  const [shaunTest, updateShaunTest] = React.useState(['Benelli', 'Ducati', 'BMW', 'Moto Guzzi']);
+  const [selectedMoto, updateSelectedMoto] = React.useState([
+    'Benelli',
+    'Ducati',
+    'BMW',
+    'Moto Guzzi',
+  ]);
   return (
     <>
       <Box width={0.4}>
         <MultiCombobox
           label="Filter columns"
           variant="solid"
-          placeholder={`Filtered Columns (${shaunTest.length})`}
+          placeholder={`Filtered Columns (${selectedMoto.length})`}
           items={['Benelli', 'Ducati', 'BMW', 'Moto Guzzi', 'Piaggio']}
-          onChange={updateShaunTest}
-          value={shaunTest}
+          onChange={updateSelectedMoto}
+          value={selectedMoto}
           canClearAllAfter={4}
           alwaysUsePlaceholder
           renderClearAll={({ clearSelectedItems }) => {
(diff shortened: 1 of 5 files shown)


src/components/MultiCombobox/MultiCombobox.mdx
```diff
@@ -179,6 +181,83 @@ const Example5 = () => {
 <Example5 />;
 ```
 
+A MultiCombobox can always show the placeholder.
+NOTE: you likely want to also pass the `hideLabel` prop as well so you don't have both a label and a placeholder showing at the same time
```
shaun-sweet, line 185: I was considering adding an error in the render of the function that would check for this for you and alert you via the console.  I just wanted some feedback on this route first before adding these types of things

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -348,17 +358,22 @@ function MultiCombobox<Item>({
...
-                    onClick={clearSelectedItems}
-                  >
-                    Clear All
-                  </AbstractButton>
-                )}
+                {isOpen &&
+                  canClearAllAfter &&
+                  value.length >= canClearAllAfter &&
+                  (renderClearAll ? (
+                    renderClearAll(clearSelectedItems)
```
kouknick, line 365: I don't think that we have to use two different styles of the "clear selection"  button. I believe that we should just replace the old button style with the new one.

3nvi: Kind of agree here

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -311,16 +319,18 @@ function MultiCombobox<Item>({
...
-                      <Tag
-                        as="li"
-                        key={itemToString(selectedItem)}
-                        m={1}
-                        onRemove={() => removeItem(selectedItem)}
-                      >
-                        {itemToString(selectedItem)}
-                      </Tag>
-                    ))}
+                    {alwaysUsePlaceholder && Boolean(value.length)
```
3nvi, line 322: I believe that the designs are a bit different since they don't use a `label` at all. We should rethink our approach entirely, rather than adding "exception props". We have a million variations for the MultiCombobox, to the point that it's almost hard to manage. Even If we absolutely need this ASAP, then I'd suggest we add a prop that overrides the input completely rather than relying on the placeholder (which unfortunately has the wrong opacity).

Could we sync tomorrow around it during our call?

3nvi:
Thanks @shaun-sweet ! Hopefully it didn't take you long to figure out pounce 😅 I'd want to sync in tomorrow's FE call before we make any changes here, because we're making an already complex component even more complex, so we need to align with designs into how this would evolve

src/components/MultiCombobox/MultiCombobox.tsx
shaun-sweet: yes.  should we loop jim in?

src/components/MultiCombobox/MultiCombobox.tsx
shaun-sweet: ill hit up amy as well

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -396,6 +398,29 @@ function MultiCombobox<Item>({
               isOpen={isOpen && results.length > 0}
               {...getMenuProps()}
             >
+              {isOpen && canClearAllAfter && value.length >= canClearAllAfter && (
```
shaun-sweet, line 401: had to move this to be nested in the <Menu /> component so you could select it without it closing the dropdown.

mrogach2350 (approved):

src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -11,6 +11,13 @@ import { typedMemo } from '../../utils/helpers';
 import Menu from '../utils/Menu';
 import AbstractButton from '../AbstractButton';
 import ComboBoxItems from '../utils/ComboBoxItems/ComboBoxItems';
+import Icon from '../Icon';
+
+type DefaultContentProps<T> = {
```
3nvi, line 16: nit: can we define this as `Pick<MultiComboboxProps<T>, 'value' | 'removeItem' | 'itemToString'>`? Just so we don't redefine typings...

3nvi (approved):
LGTM 🎉

PR #333, opened 2022-05-02
fix: pass in new function to multicombobox prop

shaun-sweet:
### Background

Would like to change an implementation to also be able to leverage the `isOpen` state.  Need to expose it

### Changes

- expose `isOpen` to `renderContent` function

### Testing

- None needed


kouknick (approved):
LGTM

3nvi (approved):
LGTM

PR #334, opened 2022-05-05
fix: pin deps

shaun-sweet:
### Background

https://panther-labs.slack.com/archives/C02CMFPV4KH/p1651710663303199?thread_ts=1651706666.624549&cid=C02CMFPV4KH

### Changes

- Describe changes here

### Testing

- Testing steps


3nvi (commented):
Great initiative here @shaun-sweet 🎉 , although I'd want to limit the pinning to only the bare minimum set of packages that contribute to bugs.

We can chat offline around this

PR #335, opened 2022-06-06
feat(tooltip): allow position to be set

nmartin413:
### Background

It would be nice to be able to customize the tooltip position.

### Changes

Let consumers override `position`. Default to previous value.

### Testing

- `npm test`


benjamminj (approved):
✨ ✨ ✨ ✨ ✨ ✨ ✨  yussssssss

benjamminj:
@nmartin413 can we update the docs with an example too?

shaun-sweet (approved):
LGTM!

PR #336, opened 2022-06-08
fix: keep table z-indices from polluting global z-index space

benjamminj:
### Background

Noticed the following code while working in Data Explorer

This is the table code itself for the results table, note the `zIndex: 1001`

image

This is another component _not within the table_, not `zIndex: 1002`:

image

This component _outside of the table_ should not need to be aware of the z-indexing _inside_ the table. Rather, the table itself probably needs to create a new stacking context so that other parts of the app can have independent z-indexing.

If need be, this can be done on individual tables in `panther-enterprise` by wrapping a table with a `Box` creating the new stacking context. But I think the z-index bleeding outside is serious enough that we should do this in the `Table` itself. Hence this PR.

### Changes

- Fix z-indexing inside of table forcing other components outside of the table to accomodate.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -45,7 +45,17 @@ const Table = React.forwardRef<HTMLTableElement, TableProps>(function Table(
   );
   return (
     <TableContext.Provider value={contextValue}>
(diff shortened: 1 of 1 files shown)


nmartin413 (approved):

kouknick (approved):
nice

PR #337, opened 2022-06-10
Add onBlur on combobox and multi combobox fields

kouknick:
### Background

Up until now, our Combobox & MultiCombobox components were not propagating the onBlur events. This PR adds support to pass an onBlur callback that will be used by the input element of the combo boxes. This in turn will allow us to determine if a combobox input is touched or not so we can display Formik errors on our Forms  only if a field is touched.

### Changes

- Handle Focus/Blur events on comboboxes

### Testing

- Testing steps
none yet :)

### DIFF
--- a/src/components/MultiCombobox/MultiCombobox.tsx
+++ b/src/components/MultiCombobox/MultiCombobox.tsx
@@ -2,10 +2,10 @@
 import React from 'react';
 import Downshift, { DownshiftState, StateChangeOptions } from 'downshift';
 import { filter as fuzzySearch } from 'fuzzaldrin';
-import Box from '../Box';
+import Box, { NativeAttributes } from '../Box';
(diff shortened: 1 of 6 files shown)


src/components/MultiCombobox/MultiCombobox.tsx
```diff
@@ -20,7 +20,7 @@ type DefaultContentProps<T> = {
   isOpen: boolean;
 };
 
-export type MultiComboboxProps<T> = {
+export type MultiComboboxProps<T> = Omit<NativeAttributes<'input'>, 'value' | 'onChange'> & {
```
3nvi, line 23: why?

3nvi (commented):
Looks good @kouknick , can we add 1 test for each component before we merge though?

benjamminj (approved):

PR #338, opened 2022-06-18
Update downshift from version N/A to 6.0.0

jsfix-updater:
This pull request was created using the JSFIX tool (https://jsfix.live) by Coana.tech (https://coana.tech).
The JSFIX run was completed by [mR4smussen](https://github.com/mR4smussen).
It bumps downshift to version 6.0.0.

<strong>Install the [JSFIX GitHub app](https://github.com/apps/jsfix-updater) to receive other package updates from JSFIX.</strong>

***

<h3>Breaking change details</h3>

<details open><summary><strong> 🚧 - Not automatically fixed breaking changes (please check before merging)</strong></summary><blockquote class="pr-blockquote"><details open>
<summary>Affects all applications (not related to specific API usages in your code).</summary>

* [useCombobox behaviour]: When an item is highlighted by keyboard and user closes the menu using mouse/touch, the item is not selected anymore. The only selection on Blur happens using either Tab / Shift+Tab. See PR for more information: https://github.com/downshift-js/downshift/pull/1109
* [useCombobox & downshift behaviour]: When pressing Escape and the menu is open, only close the menu. When the menu is closed and there is an item selected and/or text in the input, clear the selectedItem and the inputValue. See PR for more information: https://github.com/downshift-js/downshift/issues/719
</details>

</blockquote></details><details>
<summary>Breaking changes where JSFIX found that there were no occurrences.</summary>

* Pass type to the onChange (onInputValueChange, onHighlightedIndexChange, onSelectedItemChange, onIsOpenChange) handler parameters, as specified in the documentation.
* Update TS typings for selectedItem to accept null in both useSelect and useCombobox
* Update TS typings for itemToString to accept null for the item parameter, in useSelect and useCombobox + in Downshift where this was missing
</details>

***

If you would like to provide feedback to the JSFIX developers, then please leave a comment on this pull request.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,24824 @@
 {
   "name": "pouncejs",
   "version": "0.0.0-semantically-released",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "pouncejs",
+      "version": "0.0.0-semantically-released",
+      "license": "MIT",
+      "dependencies": {
+        "@emotion/react": "^11.4.0",
+        "@emotion/serialize": "^1.0.2",
+        "@emotion/styled": "^11.3.0",
+        "@juggle/resize-observer": "^3.3.1",
+        "@reach/auto-id": "^0.15.3",
+        "@reach/dialog": "^0.15.3",
+        "@reach/menu-button": "^0.15.3",
+        "@reach/popover": "^0.15.3",
+        "@reach/tabs": "^0.15.3",
+        "@reach/tooltip": "^0.15.3",
+        "@reach/utils": "^0.15.3",
+        "@styled-system/css": "^5.1.5",
+        "@styled-system/should-forward-prop": "^5.1.5",
+        "@types/fuzzaldrin": "^2.1.4",
+        "@types/history": "^4.7.9",
+        "@types/lodash": "^4.14.171",
+        "@types/reach__menu-button": "^0.1.1",
+        "@types/react": "^17.0.15",
+        "@types/react-dom": "^17.0.6",
+        "@types/react-textarea-autosize": "^4.3.6",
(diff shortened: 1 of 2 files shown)


3nvi:
This is a breaking change, we can't accept it as is

mtorp:
Hi @3nvi,

That makes sense.

What is otherwise your impression of the pull request? Was it helpful or what you rather not have received it?

Thank you for the feedback!

3nvi:
Hey @mtorp , sorry that I closed it abruptly like that, just thought this was an automatic bot performing the change and not a person (else I wouldn't have taken such a strong action).

We already have similar PRs from Snyk and dependabot, so bumping a package is not something that we'd normally be missing. The small problem here is that it's bumping a major version which caused the tests to fail. 

Solutions would be to:

- either only bump up to a minor version [easy]
- bump a breaking change version but take into account the changelog and perform refactors [super hard to automate]

Latter case is obviously preferred, but it requires a human to intervene.

Cheers

mtorp:
> just thought this was an automatic bot performing the change and not a person

No worries! It is meant to be perceived as a bot, but it is actually (currently) closely supervised/controlled by us. 

> bump a breaking change version but take into account the changelog and perform refactors [super hard to automate]

That's kinda what this tool (JSFIX) attempts to do. We use some lightweight static analysis to scan the code, and then either automatically adapt the code to the breaking changes or, in cases where that's not possible, provide the user with some guidance on how to do it manually.

In the case of pounce, JSFIX found that no usages of downshift in the source code are affected by breaking changes apart from some behavioral changes that affect all usages of useCombobox as described in the PR text.

> The small problem here is that it's bumping a major version which caused the tests to fail.

Oh okay, we should definitely not create PRs with failing tests. I see that JSFIX registered that pounce's tests still succeed after the update, but it also only runs `npm test`, so maybe there are some tests it does not know about?

PR #342, opened 2022-08-24
[Feat] New `backgroundOpacity` prop & updated background on `ProgressBar`

3nvi:
### Background

This PR adds support for a new prop, available to all components named `backgroundOpacity`. 

This prop requires a sibling value of `backgroundColor` in order to work, so specifying a `backgroundOpacity` will only work if `backgroundColor` is also defined on the same component

Similar to how [Tailwind](https://tailwindcss.com/) implements it, we cannot rely on HTML element style parsing to "figure out" what background color an element holds. This is not only hard, but incurs a performance hit. 

### Usage

```tsx
<>
  <Box backgroundColor="red-100" backgroundOpacity={50} /> // equivalent to 50%
  <Box backgroundColor="red-100" backgroundOpacity={0.5} /> // equivalent to 50%
  <Box backgroundColor="red-100" backgroundOpacity={100} /> // equivalent to 100%
  <Box backgroundColor="red-100" backgroundOpacity={1} /> // equivalent to 100%
<>
```

### Screenshots

Screenshot 2022-08-24 at 5 02 15 PM

### Changes

- Add new prop
- Add snapshots tests for new prop

### Testing

- Snapshot
- Manually

### DIFF
--- a/src/system/utility.ts
+++ b/src/system/utility.ts
@@ -1,6 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import css, { SystemCssProperties } from '@styled-system/css';
 import { StylingProps } from './system';
+import { addOpacity } from '../utils/helpers';
+import colors from '../theme/colors';
 
 export const truncateProp = ({ truncated }: any): any => {
   if (truncated) {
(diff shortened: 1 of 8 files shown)


src/system/utility.ts
```diff
@@ -39,4 +62,7 @@ export type UtilityProps = {
 
   /** Makes the component invisible to the eye, but still readable by screen readers */
   visuallyHidden?: boolean;
+
+  /** Adds an opacity to the existing `backgroundColor` */
+  backgroundOpacity?: string | number;
```
alexmylonas, line 67: As an alternative, you could make this an enumeration `[0,5,10...]` and/or `[0.25,0.5 ... ]` more restrictive but I don't see much value on having a custom opacity i.e. `0.63`

alexmylonas (approved):
LGTM! Personally I would prefer an enumeration for opacity but I am ok to merge as is.

kouknick (approved):
Thanks for that. No more absolute positioning workarounds 😄

src/system/utility.ts
3nvi: Neither do I. I also don't want to add input validation at this level :/

benjamminj (approved):
LGTM 🎉

PR #345, opened 2022-08-29
Update Alert solid variant

eliskaachee:
## Background

Fixes an issue with solid variant of snackbar not having enough contrast.

### Updates color scheme for "solid" alert variant, which was only being used in the Snackbar component

| Before  | After |
| ------------- | ------------- |
| image | image |
| image | image |

### Adds a parameter to customize the location of the snackbar
| Before  | After |
| ------------- | ------------- |
| https://user-images.githubusercontent.com/10763539/188725823-bb08133c-7c91-4a3b-94b0-100c48801bc1.mov | https://user-images.githubusercontent.com/10763539/188225340-9a73d94c-f06a-4342-b470-5cd71ce22179.mov|

### Adds an optional variantIconColor parameter to IconButton to customize the color

| Before  | After |
| ------------- | ------------- |
| image  | image |

image

## Testing

Docs for "Alert" and "Snackbar" are up-to-date
Other uses of "Alert" and "IconButton" are not affected

### DIFF
--- a/src/components/IconButton/IconButton.mdx
+++ b/src/components/IconButton/IconButton.mdx
@@ -72,7 +72,7 @@ An IconButton can be `ghost`:
     variant="ghost"
     icon="user"
     aria-label="User"
-    variantColor=-500"
+    variantColor="teal-500"
     variantBorderStyle="circle"
     active
   />
@@ -109,5 +109,33 @@ An IconButton can have no styling:
 <SimpleGrid columns={4} gap={4}>
   <IconButton variant="unstyled" icon="user" aria-label="User" />
   <IconButton variant="unstyled" icon="user" aria-label="User" variantBorderStyle="circle" />
+</SimpleGrid>
+```
+
+An IconButton can have a colored icon:
+
+```typescript jsx
+<SimpleGrid columns={4} gap={4}>
+  <IconButton
+    icon="user"
+    variantColor="navyblue-300"
+    variantIconColor="teal-500"
+    aria-label="User"
+  />
+  <IconButton
+    variant="ghost"
+    icon="user"
+    variantColor="navyblue-300"
+    variantIconColor="teal-500"
+    aria-label="User"
+  />
+  <IconButton
+    variant="outline"
+    icon="user"
+    variantColor="navyblue-300"
+    variantIconColor="teal-500"
+    aria-label="User"
+  />
+  <IconButton variant="unstyled" icon="user" aria-label="User" variantIconColor="teal-500" />
 </SimpleGrid>
 ```
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -22,7 +22,7 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
   variantColor?: keyof Theme['colors'];
 
   /** The color of the icon */
-  variantIconColor?: keyof Theme['colors'];
+  iconColor?: keyof Theme['colors'];
 
   /** The border style of the button  */
   variantBorderStyle?: 'square' | 'circle';
@@ -41,7 +41,7 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
 export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(function IconButton(
   {
     variantColor = 'blue-400',
-    variantIconColor = 'white',
+    iconColor = 'white',
     variantBorderStyle = 'square',
     active = false,
     variant = 'solid',
@@ -53,7 +53,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(f
 ) {
   const styles = useIconButtonStyles({
     variantColor,
-    variantIconColor,
+    iconColor,
     variant,
     variantBorderStyle,
     size,
@@ -67,7 +67,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(f
       {...rest}
       ref={ref}
     >
-      <Icon color={variantIconColor} type={icon} size={size} />
+      <Icon color={iconColor} type={icon} size={size} />
     </AbstractButton>
   );
 });
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -21,6 +21,9 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
   /** The color scheme of the button */
   variantColor?: keyof Theme['colors'];
 
+  /** The color of the icon for the 'unstyled' variant */
+  variantIconColor?: keyof Theme['colors'];
+
   /** The border style of the button  */
   variantBorderStyle?: 'square' | 'circle';
 
@@ -38,6 +41,7 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
 export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(function IconButton(
   {
     variantColor = 'blue-400',
+    variantIconColor = 'white',
     variantBorderStyle = 'square',
     active = false,
     variant = 'solid',
@@ -47,7 +51,13 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(f
   },
   ref
 ) {
-  const styles = useIconButtonStyles({ variantColor, variant, variantBorderStyle, size });
+  const styles = useIconButtonStyles({
+    variantColor,
+    variantIconColor,
+    variant,
+    variantBorderStyle,
+    size,
+  });
 
   return (
     <AbstractButton
@@ -57,7 +67,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(f
       {...rest}
       ref={ref}
     >
-      <Icon type={icon} size={size} />
+      <Icon color={variantIconColor} type={icon} size={size} />
     </AbstractButton>
   );
 });
--- a/src/components/IconButton/IconButton.tsx
+++ b/src/components/IconButton/IconButton.tsx
@@ -21,7 +21,7 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
   /** The color scheme of the button */
   variantColor?: keyof Theme['colors'];
 
-  /** The color of the icon for the 'unstyled' variant */
+  /** The color of the icon */
   variantIconColor?: keyof Theme['colors'];
 
   /** The border style of the button  */
--- a/src/components/Snackbar/SnackbarContext.tsx
+++ b/src/components/Snackbar/SnackbarContext.tsx
@@ -139,7 +139,7 @@ export const SnackbarProvider: React.FC = ({ children }) => {
       return null;
     }
 
-    const getPositionProps = (position: keyof typeof positionStyles = 'bottom-left') => {
+    const getPositionProps = (position: keyof typeof positionStyles) => {
       return positionStyles[position];
     };
 
@@ -157,7 +157,12 @@ export const SnackbarProvider: React.FC = ({ children }) => {
             {...getPositionProps(position)}
           >
             {transitions
-              .filter(snackbar => snackbar.item.position === position)
+              .filter(snackbar =>
+                // handles multiple toasts in different positions
+                snackbar.item.position
+                  ? snackbar.item.position === position
+                  : position === 'bottom-left'
+              )
               ?.map(({ item: { id, ...snackbarPublicProps }, key, props: styles }) => (
                 <animated.div key={key} style={styles}>
                   <Box pt={3} ref={ref}>
--- a/src/components/Snackbar/SnackbarContext.tsx
+++ b/src/components/Snackbar/SnackbarContext.tsx
@@ -71,6 +71,15 @@ const SnackbarContext = React.createContext<SnackbarContextValue>({
   updateSnackbar: () => '',
 });
 
+const positionStyles = {
+  'top-left': { top: 0, left: 6 },
+  'top-middle': { top: 0, left: '50%', style: { transform: 'translate(-50%, 0)' } },
+  'top-right': { top: 0, right: 6 },
+  'bottom-left': { bottom: 3, left: 6 },
+  'bottom-middle': { bottom: 3, left: '50%', style: { transform: 'translate(-50%, 0)' } },
+  'bottom-right': { bottom: 3, right: 6 },
+};
+
 /**
  * A component that acts both as a state-manager and provider. It provides access to methods for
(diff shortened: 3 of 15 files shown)


src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -131,15 +131,42 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
+        case 'bottom-left':
+          return { bottom: 3, left: 6 };
+        case 'bottom-middle':
+          return { bottom: 3, left: '50%', style: { transform: 'translate(-50%, 0)' } };
+        case 'bottom-right':
+          return { bottom: 3, right: 6 };
+        default:
+          // bottom-left
+          return { bottom: 3, left: 6 };
+      }
```
benjamminj, line 160: `[change]` Since this function is pure, we should move it out of the component. This will make sure it stays pure and doesn't introduce any over-rendering / stuff like that.

---

`[suggestion]` I think we can use an object for this, we'd also get the benefit of using the keys for TS

For example, something along these lines

```ts
const positionStyles = {
  'top-left': { top: 0, left: 6 },
  // ...etc
  'bottom-left': { bottom: 3, left: 6 }
}

// default position as a default argument, either in the component props or
// the utility function
const getPositionProps = (position: keyof positionStyles = 'bottom-left') => {
  return positionStyles[position]
}

// in the component
{...getPositionProps(transitions[0]?.item?.position)}
```

benjamminj, line 160: `[change]` Since this function is pure, we should move it out of the component. This will make sure it stays pure and doesn't introduce any over-rendering / stuff like that.

---

`[suggestion]` I think we can use an object for this, we'd also get the benefit of using the keys for TS

For example, something along these lines

```ts
const positionStyles = {
  'top-left': { top: 0, left: 6 },
  // ...etc
  'bottom-left': { bottom: 3, left: 6 }
}

// default position as a default argument, either in the component props or
// the utility function
const getPositionProps = (position: keyof positionStyles = 'bottom-left') => {
  return positionStyles[position]
}

// in the component
{...getPositionProps(transitions[0]?.item?.position)}
```

src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -131,15 +131,42 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
     return ReactDOM.createPortal(
       <Flex
         position="fixed"
-        bottom={3}
-        right={6}
         flexDirection="column"
         justifyContent="center"
         alignItems="flex-end"
         zIndex={9999}
+        {...getPositionProps(transitions[0]?.item?.position)}
```
benjamminj, line 169: `[question]` Do we know what happens if multiple snackbars are fired in different positions? From this, it would seem like the position is the first one.

I'm wondering if we need to support multiple positions, i.e. you could have a snackbar from the top and one from the bottom (prob not great design, but possible)

src/components/utils/ControlledAlert/useAlertStyles.tsx
```diff
@@ -37,30 +39,81 @@ const getAlertVariantIcon = (alertVariant: AlertVariant) => {
...
+    case 'warning':
+    case 'error':
+      return 'alert-circle-filled' as const;
+    case 'info':
+      return 'info' as const;
+    case 'default':
+    default:
+      return null;
+  }
+};
```
benjamminj, line 89: `[suggestion]` Having 3 functions for the alert variant styles feels a little strange to me. I wonder if we can bundle all of the styles into a single function since the variants themselves are the same?

```ts
// I think there's keys in the theme that we could use, I don't remember the exact names though
type SolidVariantTheming = { 
  color: Theme['colors'], 
  bgColor: Theme['colors], 
  icon: Theme['icons]
}

// altername names: getSolidAlertVariantStyles, getSolidAlertVariantConfig, etc.
const getSolidAlertVariantTheming = (variant: AlertVariant) => {
  switch (variant) {
    case 'success': 
      return { color: 'green-500', bgColor: 'green-500-10', icon: 'check-circle' }
    // ...other cases
  }
}
```

Thoughts?

src/theme/colors.ts
```diff
@@ -28,6 +28,7 @@ const colors = {
   'navyblue-800': '#0F1623',
   'navyblue-900': '#0A0F1A',
 
+  'blue-400-10': '#E6EFFA', // 10% of blue-400
```
benjamminj, line 31: `[discussion]` I think instead of adding new colors in this fashion we should be using either the opacity variants that @3nvi added or perhaps a combination of the `blue-400` + `10% opacity` + `white` background.

Either that or there's a `lightenDarkenColor` utility here in pounce you may be able to use on the alert bgs to get the proper mix of color + white

benjamminj (commented):
Thanks @eliskaachee!! I have a couple comments I think we should discuss thru before I'm ready to give this a green light 😄 

Also, since this is a library, we need to do a couple extra things before this PR is ready to go:

- We need to update the `.mdx` files for the components affected with examples of the changes made
- We need to add tests for any new / changed props
- Can you add some screenshots to the PR? Will make it much easier to review 🙃

src/components/Snackbar/SnackbarContext.tsx
eliskaachee: I don't know what would happen if they were fired off from different positions. Like, I think it would be rare/weird for scenario would come up, but yeah, we sould make it explicit. Here, only snackbars of one position were coming in at a time, and I needed to access the props outside of the loop, which is why I just grabbed the first one. I'll ask Sam about that. Thanks!

src/components/utils/ControlledAlert/useAlertStyles.tsx
```diff
@@ -1,66 +1,115 @@
...
 >;
 type AlertVariant = UseControlledAlertStylesProps['variant'];
+type ThemeColor = keyof Theme['colors'];
 
-const getAlertVariantColor = (alertVariant: AlertVariant) => {
+const getSolidAlertThemes = (alertVariant: AlertVariant): VariantTheming => {
+  const theme = useTheme();
+  const getBackgroundColor = (variantColor: ThemeColor) => {
+    return lightenDarkenColor(theme.colors[variantColor], 190) as ThemeColor; // TODO: This is as close as I could get to the proper color, and it's a very yellow green...
+  };
```
eliskaachee, line 17: With old styles, where I hard-coded the color:
SideSheet
When I try to use the lightenDarkenColor (it keeps getting yellower, not lighter. When I put in 10, or 90 to match "10% Opacity", it's dark green. I think I'm using it wrong:
Everything went perfectly

benjamminj: Woah, this is interesting, sounds like `lightenDarkenColor` is _**not**_ the way to go! 😱 

I still think there may be a workaround on the color so that we don't have to add a new one to the design system, perhaps we can mix it with white or something like that. Happy to pair program tomorrow on that with ya if that helps move this forward!

src/components/IconButton/IconButton.tsx
```diff
@@ -21,6 +21,9 @@ export interface IconButtonProps extends NativeAttributes<'button'>, Pick<BoxPro
   /** The color scheme of the button */
   variantColor?: keyof Theme['colors'];
 
+  /** The color of the icon for the 'unstyled' variant */
+  variantIconColor?: keyof Theme['colors'];
```
3nvi, line 25: Hmm I'm not sure I like that the `variantIconColor` is only applied to the "unstyled" variant. This is a bit unexpected for me as a consumer of this lib.

I think we should either do it for all styles or for none :/

src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -71,6 +71,15 @@ const SnackbarContext = React.createContext<SnackbarContextValue>({
   updateSnackbar: () => '',
 });
 
+const positionStyles = {
```
3nvi, line 74: nit: we should type that using `keyof ControlledAlertProps['position']: Record<string, any>` or smth similar

3nvi (commented):
Thanks for the work!

I think I need the following to approve:

- tests on the component's changes
- a few screenshots on how your changes affect the components (before -> after)
- decision on `variantIconColor` since I don't feel it's a good idea to only apply this prop only when another prop has a specific value


Other than that it looks good!

src/components/IconButton/IconButton.tsx
eliskaachee: Fixed it! Now it applies to all variants. And I added a demo / test for it

src/components/IconButton/IconButton.tsx
```diff
@@ -47,7 +51,13 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(f
   },
   ref
 ) {
-  const styles = useIconButtonStyles({ variantColor, variant, variantBorderStyle, size });
+  const styles = useIconButtonStyles({
+    variantColor,
+    variantIconColor,
```
benjamminj, line 56: @3nvi @eliskaachee Perhaps we should name this `iconColor` since the icon color itself may be independent of the bg color of the alert

thoughts?

In a perfect world I could imagine having the `variant` prop being `success | info | error | warn | default` and all colors derive from that. Or the "color" prop being a more minimal value than every color in the theme.

But that would be a much larger change, given we already have a `variant` prop and we currently do color as `variantColor`. I'd personally advocate we name the prop `iconColor`

src/components/IconButton/IconButton.mdx
```diff
@@ -72,7 +72,7 @@ An IconButton can be `ghost`:
     variant="ghost"
     icon="user"
     aria-label="User"
-    variantColor=-500"
+    variantColor="teal-500"
```
3nvi, line 75: oh cheers!

src/components/IconButton/IconButton.tsx
3nvi: I'm indifferent here. I'd say both make sense, since we already have `variantColor`

src/components/Snackbar/SnackbarContext.tsx
3nvi: @eliskaachee would you agree here? It's a small nitpick, but I feel it would help :)

src/components/utils/ControlledAlert/ControlledAlert.tsx
```diff
@@ -104,7 +114,13 @@ const ControlledAlert = React.forwardRef<HTMLDivElement, ControlledAlertProps>(
         </Flex>
         {discardable && (
           <Box my={-2} mr={-2} ml={2}>
-            <IconButton aria-label="Discard" variant="unstyled" icon="close" onClick={onClose} />
+            <IconButton
+              aria-label="Discard"
+              variant="unstyled"
+              variantIconColor={variantBackgroundStyle === 'solid' ? 'black' : undefined}
```
3nvi, line 120: nit: that should be derived from `useAlertStyles`. All logic related to styling should belong to this hook so we have a centralized place to change stuff 😉

src/components/utils/ControlledAlert/ControlledAlert.tsx
```diff
@@ -84,14 +93,15 @@ const ControlledAlert = React.forwardRef<HTMLDivElement, ControlledAlertProps>(
                   fontSize={description ? 'large' : 'medium'}
                   flexGrow={1}
                   mr="auto"
+                  pr={variantBackgroundStyle === 'solid' ? 6 : 0}
```
3nvi, line 96: question: How come the background style changes the padding? To me, as a consumer, it doesn't make sense to have padding on the right side if my background is solid vs transparent? 

If we need that, should we perhaps tie it with another prop?

src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -131,24 +139,40 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
+      <>
+        {listOfPositions.map(position => (
+          <Flex
+            key={position}
+            position="fixed"
+            flexDirection="column"
+            justifyContent="center"
+            alignItems="flex-end"
+            zIndex={9999}
+            {...getPositionProps(position)}
```
3nvi, line 157: optional: you could simply do: 

```suggestion
            {...positionStyles[position]}
```

I don't see a reason why not 😉

src/components/Snackbar/SnackbarContext.tsx
```diff
@@ -131,24 +139,40 @@ export const SnackbarProvider: React.FC = ({ children }) => {
...
+          >
+            {transitions
+              .filter(snackbar =>
+                // handles multiple toasts in different positions
+                snackbar.item.position
+                  ? snackbar.item.position === position
+                  : position === 'bottom-left'
```
3nvi, line 164: nit: not 100% sure what happens here, but I trust you. If it's not too much trouble, add an even more detailed comment to help future readers

3nvi (commented):
Thanks @eliskaachee ! I still have a few minor comments here.

When we address the style-related comment on the `ControlledAlert` I'd be happy to approve. If timezone is an issue & it's important to merge asap, feel free to rely solely on @benjamminj's approval

src/components/utils/ControlledAlert/ControlledAlert.tsx
eliskaachee: I could give this to both -- in the design, there's some extra padding on the snackbars, so I just applied it to the variant used in Snackbars

src/components/Snackbar/SnackbarContext.tsx
eliskaachee: . . . I feel like there was a long, round about way I got to this, but yes, it looks like it can be simplified XD Thanks!

src/components/Snackbar/SnackbarContext.tsx
eliskaachee: Added it! Thanks :)

src/components/utils/ControlledAlert/ControlledAlert.tsx
eliskaachee: There are also 3 other spots in that file where variantBackgroundStyle is affecting things not related to the background, like flex-direction and margins. This might be part of a bigger refactor, or establishing what the two variants are each for.

src/components/utils/ControlledAlert/ControlledAlert.tsx
eliskaachee: FIxed it!

src/components/Snackbar/SnackbarContext.tsx
eliskaachee: Updated :)

benjamminj (approved):
Looks great and all comments have been addressed! Great work @eliskaachee!!

Ideally, can we add a test for the `Snackbar` component before we merge? It doesn't have any tests right now, so I'm not gonna block based on that, but this would be a good opportunity to increase the coverage of it.

PR #347, opened 2022-08-30
feat: add `borderOpacity` prop

3nvi:
### Background

Similar to how `backgroundOpacity` was added, this PR adds `borderOpacity` to do the same thing for border colors (rather than just background colors).

The logic is the same

### Changes

- add new `borderOpacity` prop

### Testing

- Testing steps

### DIFF
--- a/src/system/utility.ts
+++ b/src/system/utility.ts
@@ -29,13 +29,7 @@ export const visuallyHiddenProp = ({ visuallyHidden }: any): any => {
   }
 };
 
-export const backgroundOpacityProp = ({
-  backgroundOpacity,
-  backgroundColor,
-}: {
-  backgroundOpacity: number | string;
-  backgroundColor: keyof typeof colors;
-}): any => {
+export const backgroundOpacityProp = ({ backgroundOpacity, backgroundColor }: any): any => {
   if (backgroundOpacity) {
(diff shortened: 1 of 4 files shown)


src/system/utility.ts
```diff
@@ -29,27 +29,34 @@ export const visuallyHiddenProp = ({ visuallyHidden }: any): any => {
...
 };
 
-export const backgroundOpacityProp = ({
-  backgroundOpacity,
-  backgroundColor,
-}: {
-  backgroundOpacity: number | string;
-  backgroundColor: keyof typeof colors;
-}): any => {
+export const backgroundOpacityProp = ({ backgroundOpacity, backgroundColor }: any): any => {
```
alexmylonas, line 32: Why remove the explicit typings?

alexmylonas (approved):
LGTM with a minor question about the removal of explicit typings.

src/system/utility.ts
3nvi: TS was failing with 3 pages of messages that were un-debuggable. Ultimately, the external typings (the ones the developers see) remain the same. I just changed the internal ones (the one internal to the typed-system) to satisfy some weird messages.

P.S. I've checked and major libraries like `chakra-ui` do the same thing cause `@styled/system` does not play great with TS

PR #348, opened 2022-09-08
[Feat] Add `closeOnOutsideClicks` on `PopoverContent` component

3nvi:
### Background

The popover closes (by default) whenever a click outside it occurs. There are cases though, where you don't want it to close when clicks outside it occur, so we should add this capability to the component

#### Showcase

https://user-images.githubusercontent.com/10436045/189114572-95aad452-5517-4177-bc13-4d42fcf3b638.mov

### Changes

- Add new `closeOnOutsideClicks` property on `PopoverContent`
- Update & write new tests for this use case
- Update docs to add a new example

### Testing

- Unit tests

### DIFF
--- a/src/components/Popover/PopoverContent.tsx
+++ b/src/components/Popover/PopoverContent.tsx
@@ -31,9 +31,9 @@ export interface PopoverContentProps {
     | 'right-top';
 
   /**
-   * Whether the popover should hide when a click outside it occurs
+   * Whether the popover should remain open when a click outside it occurs
    */
-  closeOnOutsideClicks?: boolean;
+  persistOnOutsideClicks?: boolean;
 
   /**
    * @ignore
@@ -43,7 +43,7 @@ export interface PopoverContentProps {
 
 const PopoverContent = React.forwardRef<HTMLDivElement, PopoverContentProps>(
   function PopoverContent(
-    { alignment = 'bottom-left', children, closeOnOutsideClicks = true, ...rest },
+    { alignment = 'bottom-left', children, persistOnOutsideClicks = false, ...rest },
     forwardedRef
   ) {
     const popoverAlignment = usePopoverContentAlignment(alignment);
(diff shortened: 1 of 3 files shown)


src/components/Popover/PopoverContent.tsx
```diff
@@ -30,14 +30,22 @@ export interface PopoverContentProps {
...
   /**
    * @ignore
    */
   children: ReactNode;
 }
 
 const PopoverContent = React.forwardRef<HTMLDivElement, PopoverContentProps>(
-  function PopoverContent({ alignment = 'bottom-left', children, ...rest }, forwardedRef) {
+  function PopoverContent(
+    { alignment = 'bottom-left', children, closeOnOutsideClicks = true, ...rest },
```
benjamminj, line 46: `[nit / suggestion]` Having the default value for this prop be `true` can lead to a less ergonomic interface, we may be able to tighten it up with a slight rename, flipping the logic, defaulting the new prop to `false`

```
// Right now the only time you use the prop, you HAVE to put the value
<PopoverContent closeOnOutsideClick={false} />

// If we change the name, whenever you use the boolean prop all you gotta do
// is declare it
<PopoverContent ignoreOutsideClicks />

// Default value as `false`
<PopoverContent />
```

benjamminj (approved):
This looks great! I have one suggestion on how we can make the API more ergonomic for developers, but not a blocker IMO

src/components/Popover/PopoverContent.tsx
3nvi: Yeah I thought about it, since I have the same approach when it comes to adding props (default should be `false` to help with readability & expectations), but I couldn't find a name that made sense. `ignoreOutsideClicks` is barely ok since it doesn't fully clarify what "ignore" means.

I'll go ahead and use it though :)

3nvi: Renamed to `persistOnOutsideClicks` :)

PR #350, opened 2022-09-30
[refactor] JG - adding disableAnimation prop to Alert

jesse-garrison-panther:
### Background

This adds the option to opt out of the Alert animation. A new `disableAnimation` prop can be passed in that will prevent the `Alert` from being wrapped with a `Collapse` component.

### Testing

- Tested on the documents page to ensure the animate-in functionality was not present when the `disableAnimation` was passed in.

### DIFF
--- a/src/components/Alert/Alert.tsx
+++ b/src/components/Alert/Alert.tsx
@@ -1,12 +1,20 @@
 import React from 'react';
 import ControlledAlert, { ControlledAlertProps } from '../utils/ControlledAlert';
 import Collapse from '../Collapse';
+import omit from 'lodash/omit';
 
-export type AlertProps = Omit<ControlledAlertProps, 'open' | 'onClose'> & { onClose?: () => void };
+export type AlertProps = Omit<ControlledAlertProps, 'open' | 'onClose'> & {
+  onClose?: () => void;
+  disableAnimation?: boolean;
+};
+
+const noop = () => {};
 
 /** An Alert component is simply a container for text that should capture the user's attention */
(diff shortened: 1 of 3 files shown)


src/components/Alert/Alert.tsx
```diff
@@ -16,7 +21,9 @@ const Alert: React.FC<AlertProps> = ({ onClose, ...rest }) => {
     }
   }, [setOpen, onClose]);
 
-  return (
+  return disableAnimation ? (
+    <ControlledAlert {...rest} open={open} onClose={handleClose} />
+  ) : (
```
benjamminj, line 26: `[stylistic, non-blocker]` personally, since the condition is at the top-level and simple, I'd suggest we go with an `if` block.

```
if (disableAnimation) {
  return <ControlledAlert {...rest} open={open} onClose={handleClose} />
}

return <Collapse> // etc
```

Not a blocker though!

benjamminj (commented):
Thanks @jesse-garrison-panther!!! I think in order for this to be ready to go, can we add a test + an example in the MDX file (that way folks can manually test it in the future) 🔥

jesse-garrison-panther:
I've included a test as well as updated the mdx file to include example usage of the new prop.

src/components/Alert/Alert.tsx
jesse-garrison-panther: Done! I've also included the test and examples to the MDX file.

benjamminj (approved):

PR #351, opened 2022-10-11
fix: update Modal to allow passing of onDismiss prop

purplepanda:
### Background

Related to [https://app.asana.com/0/1201267919523621/1200485078649931/f](https://app.asana.com/0/1201267919523621/1200485078649931/f)

The `ProductConsentModal`, served when a user first logs in to Panther, may be closed by hitting "escape" or clicking outside the modal. While this is typically desired behavior for modals for accessibility reasons, in this instance, we've allowed a loophole for users to use Panther without _explicitly_ agreeing to the legal terms. In advance of our trial/Odin launch, we need to close this loophole as we will have less oversight of these users.

### Changes

- Extends the Pounce Modal to allow the `onDismiss` prop to be passed.

This is part 1 of 2 for this story. There is additional work to be done in `panther-enterprise`.

Docs for this prop: https://reach.tech/dialog/#dialog-ondismiss

### Testing

- Testing steps

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -26,6 +26,14 @@ export interface ModalProps {
 
   /** The id of the HTML node that contains the text of the modal */
   'aria-describedby'?: string;
+
+  /** Callback fired when user hits "Escape" or clicks outside the dialog.
+   *
+   * When _strictly necessary_, this can be used to prevent a user from closing
+   * the dialog, but otherwise it is important to allow users to close dialogs
+   * for accessibility reasons.
+   */
+  onDismiss?: (event?: React.SyntheticEvent) => void;
 }
 
 /**
@@ -37,6 +45,7 @@ const Modal: React.FC<ModalProps> = ({
   children,
   open,
   onClose,
+  onDismiss,
   showCloseButton,
   ...rest
 }) => {
@@ -55,7 +64,7 @@ const Modal: React.FC<ModalProps> = ({
             <AnimatedDialogOverlay
               key={key}
               isOpen={item}
-              onDismiss={onClose}
+              onDismiss={onDismiss ? onDismiss : onClose}
               style={{ overflow: 'visible', opacity: styles.opacity }}
               as={'div'}
             >


benjamminj (approved):

src/components/Modal/Modal.tsx
```diff
@@ -26,6 +26,14 @@ export interface ModalProps {
...
   /** The id of the HTML node that contains the text of the modal */
   'aria-describedby'?: string;
+
+  /** Callback fired when user hits "Escape" or clicks outside the dialog.
+   *
+   * When _strictly necessary_, this can be used to prevent a user from closing
+   * the dialog, but otherwise it is important to allow users to close dialogs
+   * for accessibility reasons.
+   */
+  onDismiss?: (event?: React.SyntheticEvent) => void;
```
3nvi, line 36: Although I understand the motivation here. Keeping both `onClose` and `onDismiss` seems counter-intuitive to me. I'd love to see a way to merge these together in a single prop. Technically, the `ESC` keypress should trigger the `onClose` but it currently bypasses it (if I'm not mistaken).

I'm unsure if that was explored but it wasn't possible, so I'm interesting to hear about it

3nvi: Actually, seeing the changes below, why don't we simply add:

```ts
 onClose: (event?: React.SyntheticEvent) => void;
```

and achieve the same thing?  Wouldn't that be the best backwards-compatible way to achieve this?

georgesimos: Agree, `onClose` will do the Job!

benjamminj: @purplepanda If we can get away with a single optional `onClose` prop, that's likely preferable and I'd suggest we revert this PR and publish a new patch version of pounce. 

If that's the case I apologize for the churn, I know this fix was more on the urgent side 🙃 

Right now the core difference b/w `onClose` and `onDismiss` is the close button—I think a case can be made that those can (should) be the same prop.

Thoughts?

purplepanda: @3nvi  @benjamminj  @georgesimos  

I disagree for a couple of reasons:

1. `onDismiss` is the prop made available by the component library we've built the modal on top of **specifically** for the purpose of overriding the ability of users to close the modal via the "escape" key or clicking outside the modal. Because this is not behavior we actually want to allow under most circumstances, I think it should remain separate from `onClose`
2. It isn't true that `onClose` and `onDimiss` would always use the same method. [The documentation provides an example.](https://reach.tech/dialog/#dialog-ondismiss)

purplepanda: Most importantly, I want every engineer asked to implement this behavior to take extra care to verify that it's strictly necessary. Although we typically avoid adding friction, I think this is a case that merits it.

3nvi: Correct me if I'm wrong here, but the only difference between those 2 props is the fact that the `onDismiss` receives an event. Other than that, for all intents and purposes, it behaves exactly as `onClose` would. I'm basing that [off of this](https://github.com/panther-labs/pounce/pull/351/files#diff-766208c965d34c3bf67b6cf8dc749036c694359ca121365e8e7e27c08454383eR67). The overriding of the Escape key doesn't come from the usage of `onDismiss` AFAICT, but from the fact that you're doing `event.preventDefault()` using the event, no?

The docs that you linked mention the `onDismiss` that you see on the left side of the assignment `onDismiss={onDismiss ? onDismiss : onClose}`, so I don't know if that doc says anything more other than "use `onDismiss` to control the closing/opening of the dialog".  What we simply did was rename the `onDismiss` that `@reach-ui` has to `onClose` so it's aligned with the rest of our components

benjamminj: Thanks @purplepanda for the extra clarification 🙃 

Circling back around to this: I do think there are worthwhile discussion points on all sides, but I think here in this GitHub issue is probably not the most effective place to keep driving consensus.

I moved next week's frontend sync to Tuesday due to quality symposium, are we all ok to put this on the agenda for that meeting?

purplepanda: @benjamminj I'm not sure this merits a discussion in the frontend sync.

I think what @georgesimos and @3nvi are essentially arguing is that utilizing the `onDismiss` prop is a case of over-engineering.

@georgesimos made the point in my other PR that we don't need to support two different actions for `onClose` and `onDismiss` and @3nvi 's comments above are similar.

I misspoke in my earlier comment that `onDismiss` is **solely** for the use of overriding the ability of users to close the modal via the "escape" key or clicking outside the modal. This is one use, but of course, an infinite number of callbacks could be passed for an infinite number of reasons. That said, the issue I was asked to fix falls clearly into the use-cases for this prop.

And I don't think I'm alone in imagining that there are scenarios where `onClose` and `onDismiss` do different things because the writers of this component library provided an example.

While I agree that at this very moment we are not being asked to support this infinite number of possibilities, the `onDismiss` prop is documented OOTB functionality. It also costs nothing and does no harm to extend our modal to use it. I'm honestly surprised about the amount of debate an OOTB prop is inspiring.

3nvi: > is a case of over-engineering.

I wouldn't say over-engineering per say. I'd say that we're simply adding 2 different ways to do the same thing, while we could have one.

> And I don't think I'm alone in imagining that there are scenarios where onClose and onDismiss do different things because the writers of this component library provided an example.

I'm not sure I fully understand that. `reach-ui` only has the `onDismiss` prop and doesn't use or recommend something along the lines of `onClose`. What we had before was an `onClose` that was directly mapped to `onDismiss` and the reason for this rename was alignment with other components in pounce.

I'm not opposed to keeping `onDismiss`, but I'd recommend removing `onClose`. I can't see a reason why someone would need both props in this component. If you can give me one, then I can be swayed into keeping it. This is not something that should block any work, I just want to make sure we align in the way we view things so that we don't end up in a similar "debate" down the line.

> It also costs nothing and does no harm to extend our modal to use it. I'm honestly surprised about the amount of debate an OOTB prop is inspiring.

While it doesn't seem like it, we do try and keep pounce straightforward for consumers 😄 The reason we're talking about this is that we're adding a prop that could be a source of confusion. For me **subjectively** merging these 2 props into 1 eliminates confusion and that's something that I care about.

I would want to talk about it during the FE sync if you don't mind, not because having an extra prop is the end of the world, but purely because I want to make sure that as a team we view things under a similar spectrum. For example, I care about intuitiveness, but perhaps that's something that we shouldn't care as much about. That's perfectly fine! Let's chat and see :)

PR #354, opened 2022-10-25
fix: set z-index on modal overlay to fix conflict with side navigation overlay

purplepanda:
### Background

Relates to [https://app.asana.com/0/1202391056484782/1203229283493663/f](https://app.asana.com/0/1202391056484782/1203229283493663/f)

While conducting usability testing for the Panther Guide, a number of "paper cuts" in our user experience surfaced. This PR fixes a conflict between the modal component's overlay and the side navigation.

### Changes

- Increases the z-index of the modal overlay to ensure it will sit on top of the side navigation. Expected behavior is that the modal will close when a user attempts to interact with the side nav.
- Unit tests

Coverage report:
Screen Shot 2022-11-08 at 10 42 42 AM

### Testing

- Testing steps


src/components/Modal/Modal.tsx
```diff
@@ -65,7 +65,7 @@ const Modal: React.FC<ModalProps> = ({
               key={key}
               isOpen={item}
               onDismiss={onDismiss ? onDismiss : onClose}
-              style={{ overflow: 'visible', opacity: styles.opacity }}
+              style={{ overflow: 'visible', opacity: styles.opacity, zIndex: 100 }}
```
3nvi, line 68: nit: can we make this configurable through a prop like `backdropZIndex` (or smth similar) and default this value (or even `1000` like bootstrap) ?

Different apps might have a different needs for a `zIndex` value. Thoughts?

purplepanda: Yes, we can definitely make this configurable.

src/components/Modal/Modal.tsx
```diff
@@ -34,6 +34,9 @@ export interface ModalProps {
    * for accessibility reasons.
    */
   onDismiss?: (event?: React.SyntheticEvent) => void;
+
+  /**  The z-index to be applied to the modal overlay. Overrides the default value of 1000  */
```
3nvi, line 38: This way our docs pick up the default value 😉 
```suggestion
  /**  
     * The z-index to be applied to the modal overlay  
     * @default 1000
     */
```

3nvi (approved):
Hey @purplepanda ! It looks good but I'd love to see a unit test that makes sure this prop works as expected

src/components/Modal/Modal.tsx
```diff
@@ -47,6 +53,7 @@ const Modal: React.FC<ModalProps> = ({
   onClose,
   onDismiss,
   showCloseButton,
+  overlayZIndex,
```
georgesimos, line 56:
```suggestion
  overlayZIndex = 1000,
```

src/components/Modal/Modal.tsx
```diff
@@ -65,7 +72,11 @@ const Modal: React.FC<ModalProps> = ({
               key={key}
               isOpen={item}
               onDismiss={onDismiss ? onDismiss : onClose}
-              style={{ overflow: 'visible', opacity: styles.opacity }}
+              style={{
+                overflow: 'visible',
+                opacity: styles.opacity,
+                zIndex: overlayZIndex || 1000,
```
georgesimos, line 78: Nit, could we use a default value at the params level?

src/components/Modal/Modal.tsx
```diff
@@ -34,6 +34,12 @@ export interface ModalProps {
    * for accessibility reasons.
    */
   onDismiss?: (event?: React.SyntheticEvent) => void;
+
+  /**  
+     * The z-index to be applied to the modal overlay  
+     * @default 1000
```
georgesimos, line 40: Nit, the default looks undefined with the current implementation!

georgesimos (approved):
Thanks, LGTM!

kouknick (approved):
👍

PR #357, opened 2022-11-17
feat: allow breadcrumbs to be truncated when more than 3 items exist

purplepanda:
### Background

Relates to the [Add Breadcrumb to Top Navigation Bar](https://app.asana.com/0/1202391056484782/1203253677340404/f) story.

In order to move the breadcrumbs to the Top Navigation, we need to plan for the possibility that the length of the breadcrumb component will exceed the space available. [Design has advised we update the breadcrumbs to be truncated when this occurs.](https://www.figma.com/file/A0FIgEQziIb0WZAd8Tf7Tj/04-Detections-%E2%9C%85?node-id=3027%3A134832&t=BhKCXFSaThGWm2r2-1)

### Changes

- Describe changes here

### Testing
Screen Shot 2022-11-16 at 7 21 49 PM

### Screenshots
Screen Shot 2022-11-16 at 7 31 08 PM

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.mdx
+++ b/src/components/Breadcrumbs/Breadcrumbs.mdx
@@ -11,10 +11,39 @@ const items = [
     href: '/#/Flex',
   },
   {
-    children: <MyCustomTitle />,
-    href: '/#/MyCustomTitle',
+    children: 'Third page',
+    href: '/#/Cell',
   },
 ];
 
 <Breadcrumbs items={items} />;
+```
+
+Breadcrumbs can be truncated when more than 3 items exist:
+
+```typescript jsx
+const items = [
+  {
+    children: 'First page',
+    href: '/#/Box',
+  },
+  {
+    children: 'Second page',
+    href: '/#/Flex',
+  },
+  {
+    children: 'Third page',
+    href: '/#/Modal',
+  },
+  {
+    children: 'Fourth page',
+    href: '/#/Checkbox',
+  },
+  {
+    children: 'Fifth page',
+    href: '/#/Cell',
+  },
+];
+
+<Breadcrumbs items={items} truncate />;
 ```
--- a/src/components/Breadcrumbs/Breadcrumbs.test.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.test.tsx
@@ -1,0 +1,85 @@
+import React from 'react';
+import Breadcrumbs from './index';
+import { renderWithTheme } from '../../../jest/utils';
+
+const mockCrumbs = [
+  {
+    children: 'First page',
+    href: '/',
+  },
+  {
+    children: 'Second page',
+    href: '/secondpage',
+  },
+  {
+    children: 'Third page',
+    href: '/thirdpage',
+  },
+  {
+    children: 'Fourth page',
+    href: '/fourthpage',
+  },
+  {
+    children: 'Fifth page',
+    href: '/fifthpage',
+  },
+  {
+    children: 'Sixth page',
+    href: '/sixthpage',
+  },
+  {
+    children: 'Seventh page',
+    href: '/seventhpage',
+  },
+];
+const truncatedCrumbs = mockCrumbs.slice(1, -2);
+const visibleCrumbs = mockCrumbs.filter(index => {
+  return truncatedCrumbs.indexOf(index) === -1;
+});
+
+describe('Breadcrumbs', () => {
+  it('renders all items without truncation', () => {
+    const { queryAllByRole, queryByLabelText, queryByText } = renderWithTheme(
+      <Breadcrumbs items={mockCrumbs} />
+    );
+
+    // verify expected breadcrumbs and chevrons exist
+    mockCrumbs.forEach(crumb => {
+      expect(queryByText(crumb.children)).toBeTruthy();
+    });
+    expect(queryAllByRole('presentation')).toHaveLength(mockCrumbs.length - 1);
+
+    // verify truncated menu does not exist
+    expect(queryByLabelText('Truncated breadcrumb list toggle')).toBeNull();
+  });
+
+  it('renders breadcrumbs with truncation', () => {
+    const { queryAllByRole, queryByLabelText, queryByText } = renderWithTheme(
+      <Breadcrumbs items={mockCrumbs} truncate />
+    );
+
+    visibleCrumbs.forEach(crumb => {
+      expect(queryByText(crumb.children)).toBeTruthy();
+    });
+    truncatedCrumbs.forEach(crumb => {
+      expect(queryByText(crumb.children)).toBeNull();
+    });
+    expect(queryAllByRole('presentation')).toHaveLength(3);
+    expect(queryByLabelText('Truncated breadcrumb list toggle')).toBeTruthy();
+  });
+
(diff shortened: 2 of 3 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+
+  return (
+    <React.Fragment>
+      {/* first crumb */}
+      <BreadcrumbLink crumb={firstItem} {...rest} />
+
+      {/* truncated crumbs */}
+      <Popover>
+        {() => (
+          <Flex alignItems="center">
```
3nvi, line 84: This needs to be a `ul` as well, with each children all being `li` elements (including the button that needs to be wrapped in a `<li>` element 😉 )

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
 export interface BreadcrumbProps extends Pick<LinkProps, 'as'> {
   /** A list of `BreadcrumbsItem` objects ( `{href,children}` ) that will construct the Breadcrumb */
   items: BreadcrumbItem[];
+  /** Truncates component if breadcrumb list is too long.
+   * At least 4 breadcrumbs must exist for truncation to occur.
+   */
+  truncate?: boolean;
```
3nvi, line 25: since this is defined at pounce level (which means that it's generic), can we make it more dynamic? Could we rename it to `maxDisplayedItems` with a default value of `Infinity` (meaning "no truncation")? 

Then if you say `maxDisplayedItems={5}` you would make sure that you truncate accordingly 😉

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+      {/* first crumb */}
+      <BreadcrumbLink crumb={firstItem} {...rest} />
+
+      {/* truncated crumbs */}
+      <Popover>
+        {() => (
+          <Flex alignItems="center">
+            <PopoverTrigger
+              as={Button}
+              id="truncated-breadcrumb-list"
```
3nvi, line 87: nit: do we need this ID? If we don't use it I'd remove it just in case there's a duplicate element in the same page in another component

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+              variantColor="transparent"
+              padding={0}
+              height="auto"
+              aria-label="Truncated breadcrumb list toggle"
```
3nvi, line 91: nit: `aria-label` need to describe the call to action. We can perhaps rename that to `Toggle additional breadcrumbs`

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    mockCrumbs.forEach(crumb => {
+      expect(queryByText(crumb.children)).toBeTruthy();
+    });
+    expect(queryAllByRole('presentation')).toHaveLength(mockCrumbs.length - 1);
+
+    // verify truncated menu does not exist
+    expect(queryByLabelText('Truncated breadcrumb list toggle')).toBeNull();
+  });
+
+  it('renders breadcrumbs with truncation', () => {
```
3nvi, line 56: Can we extend this test to make sure that the additional breadcrumbs are displayed when you click once and hidden when you click again?

src/components/Breadcrumbs/Breadcrumbs.mdx
```diff
@@ -11,10 +11,39 @@ const items = [
...
+    children: 'Fourth page',
+    href: '/#/Checkbox',
+  },
+  {
+    children: 'Fifth page',
+    href: '/#/Cell',
+  },
+];
+
+<Breadcrumbs items={items} truncate />;
```
3nvi, line 48: 👍

3nvi (commented):
Thanks for the work @purplepanda! I have some small comments here & there, but the overall work looks good! It's down to some small HTML semantics for me

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+            >
+              ...
+            </PopoverTrigger>
+            <PopoverContent alignment="bottom-right">
+              <Card as="ul" my={1} px={5} pt={5} shadow="dark300" minWidth={180}>
```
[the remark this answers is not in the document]
purplepanda: The `Popover` component appears to handle this situation already: 
Screen Shot 2022-11-17 at 11 55 21 AM

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+const BreadcrumbLink: React.ElementType = ({
+  crumb,
+  isLastBreadcrumb = false,
+  showChevron = true,
+  ...rest
+}: IBreadcrumbLink) => {
+  return (
+    <Flex
+      key={crumb.href}
+      as="li"
```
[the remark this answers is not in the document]
purplepanda: Thank you for pointing this out because this element should always be the child of a list. I've removed `<Card key={item.href} minWidth={180} pb={5}>` to make this true.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,48 +19,135 @@ export interface BreadcrumbItem {
...
+        variant="neutral"
+        border="none"
+        truncated
+        maxWidth="450px"
+        data-active={isLastBreadcrumb ? true : undefined}
+        {...rest}
+      >
+        {crumb.children}
+      </Link>
+      {!isLastBreadcrumb && showChevron && <BreadcrumbChevron />}
```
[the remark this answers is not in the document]
purplepanda: It's used after every breadcrumb link that isn't the last breadcrumb and isn't truncated into the popover menu. I believe this is the most efficient place to conditionally render this element.

src/components/Breadcrumbs/Breadcrumbs.tsx
purplepanda: I don't think this need to be a `ul`. The card wrapping the breadcrumbs links in this popover is a `ul`. However, I do think this needs to be a `li`.

src/components/Breadcrumbs/Breadcrumbs.tsx
shaun-sweet: i would assume we would prefer for the truncation to be set and not configurable so it is consistent across the application usage

shaun-sweet (approved):
looks good.  excited to see how this looks in the main app with the breadcrumbs up top in the nav area!

src/components/Breadcrumbs/Breadcrumbs.tsx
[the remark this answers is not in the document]
3nvi: 👍

PR #359, opened 2022-12-02
feat: enhance tooltip to accept a custom wrapper

georgesimos:
### Background

The tooltip component is a bit limited, so I introduced the content wrapper that will enhance the tooltip capability to modify the content container, e.g. background colour or adding an arrow!
Screenshot 2022-12-02 at 2 38 34 PM

### Changes

- Add a contentWrapper property
- Keep the existing as a default

### Testing

- Testing locally

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -7,9 +7,20 @@ import Box from '../Box';
 
 const AnimatedTooltipPopup = animated(TooltipPopup);
 
+const DefaultWrapper: React.FC = ({ children }) => {
+  return (
+    <Box borderRadius="medium" bg="navyblue-300" p={4} m={2} fontSize="small" boxShadow="dark250">
+      {children}
+    </Box>
+  );
+};
+
 export interface TooltipProps {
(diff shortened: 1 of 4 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -7,20 +7,30 @@ import Box from '../Box';
...
+  );
+};
+
 export interface TooltipProps {
-  /** The string or HTML that the tooltip will show*/
+  /** The string or HTML that the tooltip will show */
   content: string | React.ReactElement;
 
+  /** The wrapper that contains the contnet */
+  contentWrapper?: React.ElementType;
```
3nvi, line 23: optional: I'd rename that to just `wrapper`

3nvi (commented):
@georgesimos It looks fine, but before I approve it, can we add a test around this new capability?

src/components/Tooltip/Tooltip.tsx
georgesimos: Done!

kouknick (approved):
LGTM!

PR #365, opened 2022-12-20
[feat] update combobox & multicombobox Menu to be more dynamic

alexmylonas:
### Background

We want the dropdown of the Comboboxes to be more dynamic and not limited to the parent component, specifically when rendering with lengthy options.

In addition, the location of the parent component should designate where that dropdown should expand to.

### Changes

- Passing a ref to the `Menu` component 
- Visual change for dropdown to feel disconnected from the parent
- Checking parent page location to decide where the dropdown should be expanded to.

### Testing

- Added testing cases for `MDX` files and validated that old instance work as indented 

### Mock
Screenshot 2022-12-20 at 6 57 36 PM

### Screencast

https://user-images.githubusercontent.com/14179917/208721905-24eda29b-c369-4093-8a51-db4dd86fe2f4.mov


src/components/utils/Menu/Menu.tsx
```diff
@@ -9,16 +9,39 @@ interface MenuProps {
   maxHeight: number;
 }
 
+function getOffset(el: HTMLElement) {
+  const rect = el.getBoundingClientRect();
+  return {
+    isLeft: rect.left < window.innerWidth / 2,
+    isBottom: rect.top + window.screenY > window.innerHeight / 2,
```
kouknick, line 16: It seems like we dont use this variable do we really need it?

src/components/utils/Menu/Menu.tsx
```diff
@@ -9,16 +9,39 @@ interface MenuProps {
...
+  };
+}
+
 const Menu = React.forwardRef<HTMLElement, MenuProps>(function Menu(
   { children, isOpen, maxHeight, ...rest },
   ref
 ) {
+  const [positions, setPositions] = React.useState<{
+    top: string | number | null;
```
kouknick, line 25: Should we remove this variable and just store the right?

kouknick (commented):
Great job @alexmylonas! This was tricky.

src/components/utils/Menu/Menu.tsx
alexmylonas: We don't I started experimenting with rendering the combobox above but it took way more time than expected.

Removed it.

src/components/utils/Menu/Menu.tsx
alexmylonas: I removed the `top`but I kept the object because I think it will be a fast followup

src/components/utils/Menu/Menu.tsx
```diff
@@ -27,20 +48,20 @@ const Menu = React.forwardRef<HTMLElement, MenuProps>(function Menu(
...
+            borderColor="blue-400"
             borderBottomLeftRadius="medium"
             borderBottomRightRadius="medium"
             backgroundColor="navyblue-300"
             zIndex={10}
             position="absolute"
-            width={1}
+            minWidth="100%"
+            width="max-content"
+            maxWidth="40vw"
```
alexmylonas, line 61: Better use pixels.
```suggestion
            maxWidth="800px"
```

src/components/Tooltip/Tooltip.tsx
```diff
@@ -22,18 +21,19 @@ export interface TooltipProps {
   /** The wrapper that contains the contnet */
   wrapper?: React.ElementType;
 
-  /** The position of the tooltip relative to the content */
-  position?: Position;
+  /** The alignment of the tooltip relative to the content */
```
3nvi, line 24: made sure this was never used

alexmylonas (commented):
Did a quick user testing and LGTM!

3nvi (approved):

PR #375, opened 2022-12-21
Migrate from CircleCI to GitHub Actions

TDaglis:
### Background

Migrate CI from CircleCI into GitHub actions

Closes https://app.asana.com/0/1202423273325597/1203272741863895/f

### Changes

- Add workflow for GitHub actions

TODO:
- [ ] Add NPM_TOKEN secret - not sure where it is needed
- [ ] Add SURGE_LOGIN secret - needed for the docs step
- [ ] Add SURGE_TOKEN secret - needed for the docs step


.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+  push:
+
+jobs:
+  audit_and_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.2.0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
alexmylonas, line 13: I think you need to use `node-version: 16` here and below since it's what we are using locally.

georgesimos: Node version 14.18 was used in circleci!
I would also suggest version 16!

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+          cache: 'npm'
+      - name: Audit the installed modules
+        run: npm audit --production
```
3nvi, line 16: since we use a third-party tool for package auditing in Panther, I don't think we need that anymore

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install ci dependencies
+        run: npm ci
+      - name: Run pounce tests
+        run: npm test
```
3nvi, line 20: those will now run as part of our web integration test suite on Panther itself

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
+          cache: 'npm'
+      - name: Install js dependencies
+        run: npm i
+      - name: Lint project and run TS checks
+        run: npm run validate
+      - name: Build the project
+        run: npm run build
```
3nvi, line 34: no longer needed since we don't build Pounce on its own. It gets built as part of the Panther build process

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+          node-version: 14
+          cache: 'npm'
+      - name: Install js dependencies
+        run: npm i
+      - name: Generate documentation pages
+        run: npm run docs
+      - name: Setup documentation deployment
+        run: sudo npm install -g surge@latest
+      - name: Deploy documentation pages
+        run: surge --project .styleguidist --domain pouncejs.surge.sh
```
3nvi, line 64: No longer needed since Pounce, soon, won't have public docs anymore

3nvi (commented):
Given that we are going to archive Pounce in the coming days, I don't think the migration to GHA is needed. 

In addition

3nvi:
Closing as the testing & auditing is now part of Panther

